Issue #3, opened 2026-07-25
Fail the take on console errors, failed requests, and non-zero exit codes


rogvid:
Split from #1 (item 3). Cheapest high-value item in the whole proposal, and
it has no dependency on the manifest work.

## Problem

A demo that looks perfect while throwing `TypeError` on every render
currently passes review. Likewise `TerminalRecorder.run()` ignores the
command's exit code, so a failing command sails past unnoticed and the take
still "succeeds".

## Tasks

- [ ] `page.on("console")` and `page.on("pageerror")`, attributed to the beat
      that was active when they fired
- [ ] `page.on("requestfailed")` plus non-2xx responses
- [ ] `TerminalRecorder.run()` records the command's exit code
- [ ] Print a summary at the end of the take (errors, failed requests,
      non-zero exits) so it is visible without any extra tooling
- [ ] Optional `Recorder(..., strict=True)` fails the take on any console
      error or non-zero exit
- [ ] Document in `SKILL.md` — a clean-looking demo over a broken app is the
      failure mode this catches

## Acceptance

A storyboard against a page that throws on render exits non-zero under
`strict=True`, and names the beat during which the error fired.

## Note

The original item also said "surface all of it in `timeline.json`". That
bullet is deliberately dropped here — it depends on the beat event log slice
and re-lands there. This slice is worth having before the manifest exists.

Issue #2, opened 2026-07-25
Design: how does the demo video reach a reviewer on a PR?


rogvid:
**Status: open design question. Not ready to implement — no option here is
good enough yet.**

The "stop committing the mp4" half of this issue has shipped: `SKILL.md` now
commits `record.py`, `guide.md`, and stills, gitignores `demo.mp4`, and
points readers at the re-record command. That removes the history bloat
without needing any of what follows.

What remains is the part that isn't solved: **a reviewer on a PR should be
able to see the demo without checking out the branch and re-recording it.**
Today the documented answer is "drag the mp4 into the comment box by hand,"
which works and is honest, but doesn't automate.

## Why this is hard (research already done — don't redo it)

- GitHub has **no public API for comment attachments**. The endpoint behind
  drag-and-drop (`github.com/upload/policies/assets`) is internal and
  undocumented — see community discussion #29993, cli/cli#4465 and #12960,
  both closed as blocked on the platform. `gh` cannot attach a file to a
  comment.
- **GitHub only plays video it hosts itself.** An mp4 at any other URL —
  including a release asset — renders as a bare link, not a player. This
  repo's own `skills/demo-video/README.md` already hedges around this with
  "if the player doesn't load in your viewer."
- **Images from external URLs do render inline**, proxied through camo.
- **Release assets sit outside git history** and `gh release upload` is
  documented and stable.

## Options considered, none satisfying

| Option | Gets a player? | Cost |
|---|---|---|
| Manual drag-and-drop (today) | yes | not automatable |
| Release asset + GIF embed + mp4 link | no, GIF only | asset sprawl, GIF size ceiling |
| Internal upload endpoint (cf. `gh-image`) | yes | undocumented; breaks without notice, in CI, on someone else's PR |
| Actions artifact | no | zipped, login-walled, 90-day expiry |
| External host (S3/R2) | no | credentials + infra in a drop-in skill |
| Orphan/`gh-pages` branch | no | still git objects, just relocated |

The GIF route also can't carry a whole demo — a 30–60s recording at a
watchable size lands well past camo's proxy ceiling, so it would have to be
a short highlight loop rather than the full story.

## Open questions before this is pickable

- [ ] Is a still strip in the PR comment actually sufficient? The skill
      already captures a still at every narrated beat, and its own review
      process reads frames rather than watching video. A captioned strip may
      be the better review artifact, making the video a download rather than
      something that needs to render at all.
[...]

Issue #1, opened 2026-07-25
Emit machine-readable evidence alongside the video, so demos can act as a CI review gate


rogvid:
`demo-video` already treats the video as a scripted artifact rather than a
screen capture — the storyboard is the source of truth, stills are read
instead of watched, and a context-free agent reviews extracted frames. That
design is right and most of what follows builds directly on it.

The gap is that the *only* durable output is pixels. Claude and most
frontier models take images, not video, so any reviewing agent must
decompose the mp4 back into frames and re-infer application state that the
recorder already knew exactly. That round trip is lossy and expensive, and
it's unnecessary: the recorder is driving the page, so it can emit the
evidence directly.

Goal: same run, two artifacts. An mp4 for humans, a structured manifest for
agents. That's what makes the skill usable as a PR review gate rather than
only as documentation tooling.

Items below are independent and can be split into separate issues; they're
grouped here because they share the same underlying change (a beat-level
event log inside `_DemoBase`).

---

## 1. Emit a timeline manifest

**Problem:** `guide.md` is written by hand next to `record.py` and drifts as
soon as the storyboard changes.

**Proposal:** emit `timeline.json` and `timeline.md` on clean exit, as a
byproduct of recording rather than a parallel document.

The machinery already exists in `helpers/demo_recording/core.py`:
`_caption_hold` computes hold durations, `_start_line`/`_finish_line`
bracket narration, `shot()` knows still names, `media_duration` gives the
total. It needs a list to append to.

Per beat: `index`, `t_start`, `t_end`, `caption`, `verb`, `selector`,
`still` (if any), `segment`.

- [ ] Add a beat event log to `_DemoBase`
- [ ] Write `timeline.json` on `__exit__`
- [ ] Render `timeline.md` from it (human-readable, embeds stills)
- [ ] `stitch()` merges per-segment timelines with corrected offsets

**Acceptance:** `timeline.json` beat timestamps land within ~100ms of the
corresponding caption appearing in `demo.mp4`.

---

## 2. Capture agent-readable evidence per beat

**Problem:** a reviewing agent infers DOM state from pixels. The recorder
has the real thing.

**Proposal:** at each caption/spotlight, capture alongside the frame:

- `page.accessibility.snapshot()` scoped to the spotlight target — semantic,
  compact, stable across restyling; preferred over raw HTML
- `outerHTML` of the spotlight target
- For `TerminalRecorder`: the ANSI-stripped rendered screen (`_screen()`
  already returns this)

[...]

Issue #14, opened 2026-07-25
demo_recording is Unix-only even for web-only demos


rogvid:
Found while writing the smoke test (branch `wave0-test-harness`). Not fixed
there on purpose — that PR is forbidden from touching the recorder modules
while #3-#12 are queued against them.

`skills/demo-video/helpers/demo_recording/__init__.py` imports both recorders
eagerly:

```python
from .terminal import TerminalRecorder
from .web import Recorder
```

`terminal.py` imports `fcntl`, `pty`, and `termios` at module scope, none of
which exist on Windows. So `from demo_recording import Recorder` — the
*web* recorder, which is pure Playwright and has no PTY dependency — fails at
import time on Windows with `ModuleNotFoundError: No module named 'fcntl'`.

SKILL.md is careful to scope the Unix-only constraint to terminal recording
("Terminal recording is **Unix-only** (it uses a PTY)"), so the code is
stricter than the documentation promises.

## Impact

Low today — nobody has reported recording from Windows, and CI is Linux. It
does mean `tests/smoke` has to declare itself Unix-only in full rather than
just skipping its terminal take, which it currently documents.

## Fix

Make the terminal import lazy in `__init__.py` (a module `__getattr__`, or a
guarded import that re-raises a clear "TerminalRecorder needs a Unix PTY"
message on attribute access). Either way `Recorder` should import cleanly
anywhere Playwright runs.

rogvid:
**blocks-v1** (#62). Straightforward adoption blocker: `from demo_recording import Recorder` — the *web* recorder, pure Playwright, no PTY — fails at import on Windows because `__init__.py` eagerly imports `terminal.py`, which imports `fcntl`. If one person on the team is on Windows they cannot record a web demo at all. Lazy import; small.

PR #15, opened 2026-07-25
Add a fixture app, a smoke test, and CI for the demo-video recorders

rogvid:
Foundation PR for the #3–#12 queue. The recorders had 1,344 lines of code,
zero tests, no CI, and no committed app to record against — the showcase
storyboard pointed at a `.scratch/` server that is gitignored and long gone,
so a fresh clone or worktree could not verify a recorder change at all.

## What lands

- `tests/fixture/index.html` — single-file dashboard, no deps, no build.
  Stable ids for recording, deterministic (no RNG, no clock, no animation).
  Two inert hooks for queued work: `?console-error=1` (#3) and `?secret=1`
  rendering `sk-live-FAKE0000000000000000` (#4).
- `tests/smoke` — PEP 723 uv script per `script-conventions`. Serves the
  fixture, records a web take and a terminal take, asserts mp4 exists with a
  sane duration and that stills are non-trivial. `--web-only` /
  `--terminal-only` to narrow.
- `.github/workflows/ci.yml` — `lint`, `typecheck`, `secrets`, `smoke`.
  Uploads the recording as an artifact on failure.
- `ruff.toml`, `mypy.ini`, `.gitleaks.toml`.

Nothing under `skills/` is touched, deliberately: ten queued PRs edit
`core.py`, `web.py` and `terminal.py`, and any change here would conflict
with all of them.

## Verified locally

Smoke test passes end to end (web 14.0s/208kB, terminal 10.3s/113kB, five
stills), plus both narrowing flags, an argparse rejection, and a seeded
failure path. `ruff check`, `ruff format --check`, `mypy`, and `shellcheck`
all clean. Fixture hooks confirmed with a throwaway Playwright script.

## NOT verified

**The CI workflow has never run.** This PR is its first execution. Action
resolution, `playwright install --with-deps` on the runner, gitleaks' actual
verdict on the fixture literal, and whether the duration windows hold on a
cold runner are all unknown until the checks report here.

## Known compromises

Tracked as a ratchet checklist in #13:

- `ruff format` excludes the three recorder modules — reformatting them
  collides with every queued PR. Delete the exclusion once the queue drains.
- `mypy.ini` sets `strict_optional = False`, which is the one knob that
  clears the errors caused by `_env()`'s return type. The real fix is
  `@overload` on `_env`.
- E501 unselected; every long line is inside an embedded JS/CSS literal.

Also filed: #14, `demo_recording` is Unix-only even for web-only demos
because `__init__.py` eagerly imports the PTY-based terminal module.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,12 +63,18 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
+      # What gitleaks scans depends on the event: for `pull_request` it is the
+      # commits in the PR, for `push` the pushed range, and only for a manual
+      # or scheduled run is it the whole repository. fetch-depth: 0 is what
+      # makes that last case possible — without it a full scan has no history
+      # to read.
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 0 # gitleaks scans history, not just the checkout
+          fetch-depth: 0
 
-      # Config is .gitleaks.toml at the repo root: the default ruleset plus a
-      # single allowlist entry for the fixture's fake API key.
+      # Config is .gitleaks.toml at the repo root: the default ruleset plus two
+      # narrow allowlist entries (vendored xterm.js, and the fixture's fake
+      # API key).
       - uses: gitleaks/gitleaks-action@v2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,108 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  # Pinned so a tool release never turns a green branch red on its own.
+  RUFF_VERSION: "0.14.2"
+  MYPY_VERSION: "1.18.2"
+
+jobs:
+  lint:
+    name: lint (ruff, shellcheck)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+
+      - name: ruff check
+        run: uv run --with "ruff==$RUFF_VERSION" ruff check --output-format github .
+
+      - name: ruff format --check
+        run: uv run --with "ruff==$RUFF_VERSION" ruff format --check .
+
+      # shellcheck comes from PyPI rather than apt so this is the same binary
+      # a contributor gets from the same command locally.
+      - name: shellcheck the skills' shell scripts
+        run: |
+          find skills -name '*.sh' -print0 |
+            xargs -0 -r uv run --with shellcheck-py shellcheck --severity=style
+
+  typecheck:
+    name: typecheck (mypy)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+
+      # Settings live in mypy.ini and are deliberately permissive — see the
+      # comments there for what to ratchet up and when.
+      - name: mypy
+        run: uv run --with "mypy==$MYPY_VERSION" --with playwright mypy skills/demo-video/helpers/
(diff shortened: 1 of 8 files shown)


rogvid:
## Cold review: BLOCK

All four CI jobs passed on first run, including `smoke` on a cold runner
(1m22s) — the workflow itself is sound and the fixture/lint/typecheck work
is good. But fault injection found holes that would let a broken recorder
through, and ten PRs are queued behind this harness.

### Blocking

1. **Reused `--out-dir` is never cleared.** With `shot()` and `_convert()`
   both patched to produce nothing, a re-run into the same dir reports
   `smoke: PASSED` on stale artifacts. That is the loop `tests/README.md:24`
   documents. CI is safe only because `$RUNNER_TEMP` is fresh.
2. **`MIN_MP4_BYTES` cannot detect a blank video.** A flat-white 14s
   1280x720 H.264 is 23,643 bytes — over the floor, mid-duration-window. A
   genuinely blank recording passed both mp4 assertions at 48kB; only the
   PNG floor caught it, by 746 bytes. No threshold separates blank from real
   (terminal takes are 108-116kB), so this needs a content assertion.
3. **No-op'ing `click()` and `type_into()` passes green.** The demo records
   a static page and nothing notices. Elevated from the reviewer's
   non-blocking ranking: those verbs are exactly what #3–#12 edit, so a
   harness blind to them fails at its purpose.

### Non-blocking

- gitleaks full-history finds 2 false positives in vendored `xterm.js`
  (already on main). Green on PR events only; add a `helpers/assets/` path
  allowlist. The `fetch-depth: 0` comment at `ci.yml:68` is false for
  `pull_request` events.
- The `sk-live-FAKE0{16}` allowlist entry is inert — gitleaks does not flag
  that literal with or without it. Harmless, but the comment oversells it.
- `free_port()` TOCTOU; `ruff format --check` effectively gates one file
  (consistent with the #13 ratchet); actions 1-3 majors behind.

### Verified sound

Cleanup and exit codes are correct — a mid-take `RuntimeError` leaves no
orphaned `http.server`, `headless_shell`, or PTY processes. Duration windows
are not a flakiness risk: three concurrent runs varied by 0.1s against >2x
headroom, because pacing is wall-clock rather than machine-speed dependent.
Convention compliance against `script-conventions` is clean. The scope claim
(nothing under `skills/`) is true.

Sent back to the implementer.

rogvid:
## Second cold review: BLOCK

CI green on `a0f47d3` (all four jobs, smoke 1m28s). Fix #1 (stale out-dir)
confirmed real. Fix #3 works for `click`/`type_into`/`spotlight`. Fix #2
does not work.

### The contrast metric is anti-correlated

A fully blank web recording scores **61.8**. A healthy one scores **60.2**.
The blank scores higher.

`frame_contrast()` scores the whole frame, and 20-35% of every frame is the
recorder's own chrome — pastel gradient ~230 luma against a `#232334` window
at ~35. That fixed bimodal spread alone yields 60-79 regardless of what the
app renders. The 8.0/6.0 floors are unreachable by any output the recorder
can physically produce. Validation was against a synthetic flat-color mp4 —
a file the recorder cannot emit.

Terminal is blind in both axes: setting the xterm foreground to the
background colour gave 79/79/78 against a baseline of 79/78/78, `PASSED`.
Only **web stills** carry real signal (blank 0.0, healthy 15-17) because
they are raw page screenshots with no chrome.

### Also missed

- **`caption()` can be deleted entirely and both takes pass**, contrast
  identical to baseline. Captions are the headline feature; #3 and #12 both
  edit `caption()`. The bottom-20% crop that stopped the caption bar from
  inflating scores removed the only signal that varies.
- **`move_to()` no-ops and passes.** The assertion measures Playwright's own
  `mousemove` during `click()`, not the recorder's glide.
- **Duplicate stills pass** — three byte-identical screenshots satisfy every
  check.
- **A total CSS failure scores 9.97**, above the 6.0 floor (healthy 15-17).

### `fresh_take_dir()`

Bounded and symlink-safe (rmtree refuses symlinks; target survived). One
footgun: `--out-dir .` in a repo containing `web/` silently rmtrees it —
reproduced. Not blocking here, since CI uses `$RUNNER_TEMP`.

### Non-blocking

Fixture coupling makes diagnostics actively misleading — a legitimate
fixture edit reports "the interaction did not take effect" against a verb
that worked fine. `record_terminal()` raises rather than collecting, so a
terminal failure aborts before the mp4 is written and the `if: failure()`
artifact upload has nothing to upload.

### Clean

Nothing under `skills/` touched. No injected test code leaked. All pinned CI
commands pass locally. Content checks cost 0.37s.

Final fix round; remaining gaps get documented and filed rather than
iterated on further.

Issue #25, opened 2026-07-25
TerminalRecorder's background init script throws a TypeError on every take


rogvid:
Found while implementing #3 (console-error / exit-code detection), which is
what made it visible at all: every `TerminalRecorder` take logs an uncaught
`pageerror` before the first beat.

```
page_error in before the first beat: Cannot read properties of null (reading 'style')
```

## Cause

`TerminalRecorder._init_context` adds this init script:

```js
(() => { const bg = '…';
  document.documentElement.style.background = bg;
  const b = () => { if (document.body) document.body.style.background = bg; };
  if (document.body) b(); else addEventListener('DOMContentLoaded', b); })();
```

Init scripts run on document creation, and on Chromium's *initial empty
document* `document.documentElement` is null — so the first statement throws
before the `DOMContentLoaded` fallback below it is ever wired up. Reproduced
standalone against a bare Playwright context: pageerror on `about:blank`, stack
pointing at that line.

The `document.body` guard on the line under it shows the author knew the
document might not be ready; `documentElement` just did not get the same
treatment.

## Why it matters now

Not cosmetic any more. #3 adds `Recorder(..., strict=True)`, which fails a take
on any uncaught page exception — so **strict mode currently fails every
terminal take on the recorder's own bug**, regardless of what the demo does.
That makes the flag unusable for `TerminalRecorder` until this is fixed.

It also means the background paint this script exists to apply is *not* applied
on the initial document, which is the white-flash case its comment says it is
there to prevent.

## Fix

Guard it the same way the line below it is guarded, e.g. apply to
`documentElement` only when it exists and let the `DOMContentLoaded` handler
cover the rest.

Deliberately not fixed in #3: that PR is scoped to detection, and three PRs are
queued against `terminal.py`. Worth doing as a one-line commit on its own so
the strict-mode flag becomes useful for terminal demos.

## Acceptance

A `TerminalRecorder` take against a healthy program records zero issues in
`timeline.json`, and `TerminalRecorder(..., strict=True)` completes without
raising.

Issue #6, opened 2026-07-25
Emit a beat event log as timeline.json and timeline.md


rogvid:
Split from #1 (item 1), minus the `stitch()` merge which is its own issue.

**This is the enabler for beat-aligned frames, per-beat evidence, and AC
tagging.** Nothing else depends on it; those three all do.

## Problem

`guide.md` is written by hand next to `record.py` and drifts as soon as the
storyboard changes. More broadly, the only durable output of a take is
pixels, so any reviewing agent has to decompose the mp4 back into frames and
re-infer state the recorder already knew exactly.

## Proposal

Emit `timeline.json` and `timeline.md` on clean exit, as a byproduct of
recording rather than a parallel hand-maintained document.

The machinery already exists in `helpers/demo_recording/core.py`:
`_caption_hold` computes hold durations, `_start_line`/`_finish_line` bracket
narration, `shot()` knows still names, `media_duration` gives the total. It
needs a list to append to.

Per beat: `index`, `t_start`, `t_end`, `caption`, `verb`, `selector`, `still`
(if any), `segment`.

## Tasks

- [ ] Add a beat event log to `_DemoBase`
- [ ] Write `timeline.json` on `__exit__`
- [ ] Render `timeline.md` from it (human-readable, embeds stills)
- [ ] Fold in the console-error / exit-code data from the strict-mode slice
      if that has already landed

## Acceptance

`timeline.json` beat timestamps land within ~100ms of the corresponding
caption appearing in `demo.mp4`.

Issue #12, opened 2026-07-25
Acceptance-criterion tagging and a conformance reviewer


rogvid:
Split from #1 (item 8). Depends on the beat event log slice.

This is the slice that turns the skill from documentation tooling into a
review gate.

## Problem

The existing fresh-agent review asks *"is this story clear?"*. For a review
gate the question is *"does this demonstrate what the ticket asked for?"* —
a different question with a different answer. A demo can be perfectly clear
and demonstrate the wrong thing.

## Proposal

```python
rec.caption("The overdraft is rejected at submit.", ac="AC-3")
```

The manifest then emits coverage: criteria demonstrated, criteria missing,
beats with no criterion. "AC-2 and AC-5 are undemonstrated" becomes a
machine-checkable review finding rather than something a human has to notice.

## Tasks

- [ ] Optional `ac=` on `caption()` / `shot()`
- [ ] Coverage section in `timeline.json` and `timeline.md`
- [ ] Second reviewer prompt in `SKILL.md`: given acceptance criteria +
      manifest + beat frames, judge **conformance** — distinct from, and run
      alongside, the existing **comprehension** reviewer
- [ ] Document the anti-pattern explicitly: derive the storyboard from the
      ticket/ADR/RFC, **not** from the diff. A scenario generated from the
      implementation is a tautology, and it will produce a polished,
      convincing demo of a misread requirement

## Acceptance

A storyboard tagged against a ticket's acceptance criteria produces a
coverage report naming any criterion with no beat demonstrating it.

rogvid:
**blocks-v1** (#62). Re-reading this against the stated purpose: it is not an enhancement, it is the headline. The gap the skill exists to close is *adherence to user stories and ADRs* — the thing linting, type checking, e2e tests, security audits and reviewer agents all cannot verify. `ac="AC-3"` on a caption, and coverage in the manifest, is the direct answer. Hard dependency on #24: a conformance gate must not be able to derive "this beat demonstrated AC-3" from a verb that threw.

PR #27, opened 2026-07-25
Fail the take on console errors, failed requests, and non-zero exit codes

### LINKED ISSUE #3: Fail the take on console errors, failed requests, and non-zero exit codes

rogvid:
Fixes #3.

A demo that looks perfect while throwing `TypeError` on every render passed
review, and `TerminalRecorder.run()` ignored exit codes, so a failing command
sailed past unnoticed.

Wires `console` / `pageerror` / `requestfailed` / non-2xx responses, attributed
to the beat that was active when they fired (using the beat log from #6), plus
the exit status of every terminal command. All of it lands in `timeline.json`
and a new Issues section in `timeline.md`, and prints as an end-of-take summary
regardless of tooling.

`Recorder(..., strict=True)` refuses the take on `console_error`, `page_error`
or `nonzero_exit` — raised from `__exit__` **after** the mp4, stills and
timeline are written, since a broken take is the one worth looking at.

`TIMELINE_SCHEMA` stays at 1. Every change adds a key; none changes meaning or
type, which is what the contract comment permits without a bump.

Exit status is read via an OSC escape carrying `$?` embedded in the shell's
PS1, stripped from the PTY stream before xterm.js renders it — so nothing
types `echo $?` into the demo. Bash-shaped; a shell that does not expand `$?`
in its prompt yields `exit_code: null` rather than a wrong number.

## What strict does NOT catch — documented, not engineered around

Chromium writes its own console error for anything that 404s, so a strict take
fails on a missing favicon. Filtering that would mean pattern-matching
Chromium's English console wording, which is worse. `console_warning`,
`request_failed` and `http_error` are recorded but not fatal. Nothing catches
what the app swallows — a caught exception, a silently wrong render, a 200
with wrong data.

## Verification

Eleven fault injections, each watched fail: unsubscribed listeners, misattributed
beats, discarded exit status, `nonzero_exit` dropped from `STRICT_KINDS`, the
Issues section removed from `timeline.md`, an off-by-one `issue_count`, the
strict raise deleted, a strict message that does not name the beat, strict
forced on by default, the raise moved ahead of the artifact writes, and an
out-of-contract issue kind.

## Found on the way — read #25 before relying on strict mode

`TerminalRecorder`'s background init script dereferences
`document.documentElement` on Chromium's initial empty document, where it is
null, so **every terminal take throws an uncaught TypeError before its first
beat**. Pre-existing; this PR's instrumentation is what made it visible.
Consequence: `strict=True` currently fails every terminal take regardless of
the demo. Deliberately not fixed here (three PRs are queued against
`terminal.py`), and the smoke assertions are immune to it.

Unverified and documented in `tests/README.md`: the 200-issue cap, a `run()`
never followed by `wait_for_prompt()`, non-bash shells, segments, and
`console_warning` (implemented, nothing emits one).

### DIFF
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -256,25 +256,45 @@ app itself and writes what it saw into `timeline.json` as `issues`:
 | `http_error` | a response with status ≥ 400 (3xx redirects are normal) | no |
 | `nonzero_exit` | a `TerminalRecorder` `run()` whose command failed | yes |
 
-Every issue is **attributed to the beat that was running when it fired** —
+Each issue is **attributed to the beat that was running when it fired** —
 `beat` (an index into `beats`), plus the beat's `verb` and `caption` copied
 alongside so the list reads on its own. "The take broke" is not a bug report;
 "the take broke during `click('#refresh')`, under the caption *Refresh reloads
 it*" is. `timeline.md` gets an **Issues** section saying the same thing in
 prose, so a reviewer reading the PR sees it without opening the JSON.
 
+**`beat` is `null` when no beat can honestly claim the problem**, and that is a
+real answer rather than a gap. Playwright hands the recorder page events only
+while it is being called, so the naive reading — blame the most recently
+started beat — invents attributions in both directions: an error thrown during
+a three-second `hold()` would be blamed on the beat *after* the hold and quoted
+under a caption that had not appeared yet. Holds therefore pump events as they
+wait, and anything still ambiguous — a problem surfacing between two verbs,
+or after a long stretch where nothing reached Playwright — records `beat: null`
+instead of a confident guess. Trust `beat`; `t` is when the problem was
+*observed*, which can lag when it happened.
+
 Nothing has to be asked for: **a summary prints on stderr at the end of every
 take**, listing each problem and its beat, or saying plainly that there were
 none.
 
-`TerminalRecorder.run()` additionally records `exit_code` on its beat. The
-shell reports it through an invisible escape sequence in its own prompt, which
-the recorder strips before the terminal ever renders it — so the status is
-known without typing `echo $?` into the demo. It arrives when the prompt comes
-back, which is normally during the `wait_for_prompt()` that follows; a `run()`
-that is never waited on can end the take with `exit_code: null`. Non-bash
-shells that do not expand `$?` in their prompt also leave it null, rather than
-wrong.
+`TerminalRecorder.run()` additionally records `exit_code` on its beat, and
+`timeline.md` gets an `exit` column when any beat has one. The shell reports
+the status through an invisible escape in its own prompt, carrying `$?` and
+bash's command number, which the recorder strips before the terminal ever
+renders it — so the status is known without typing `echo $?` into the demo.
+
+The command number is what makes it trustworthy. The shell prints a prompt at
+startup before any command, and reprints one for an empty Enter or a Ctrl-C, and
+each of those reports a status belonging to no command; the number is what tells
+them apart. Two `run()`s with no wait between them queue, and each status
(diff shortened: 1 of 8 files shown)


rogvid:
## Cold review: BLOCK

CI green on all four jobs. The OSC stripping survived every leak attempt, and
schema compatibility verified clean.

### Blocking

1. **`exit_code` is silently wrong, not null.** Two reachable paths — an
   unwaited `run()` before a slow command, and a shell-startup race
   reproducible with a 1.2s-delayed shell. `(exit 9)` records as `0`, no
   issue logged, `strict=True` passes. This reinstates the exact bug #3
   exists to fix, and the "null rather than wrong" contract is asserted in
   three places.
2. **Attribution names a beat that had not started when the error fired.**
   An error 1s into a 3s `pause` is recorded 2s late, attributed to the
   *following* beat, and stamped with a caption that appeared after it.
   `self._beats[-1]` is unbounded in both directions — the smoke's own probe
   fetches attribute to an already-closed beat. A confidently wrong beat
   index is worse than `beat: null`, and #12 will gate on this.
3. **The harness cannot see false positives.** Injecting `_on_response` to
   report every healthy 2xx as a fatal `console_error` yields
   `smoke: PASSED`. A recorder that refuses every strict take of every
   working app passes the whole suite. Every issue row is "at least one
   matches" and nothing asserts a healthy take records nothing — all eleven
   injections probe one direction. **#25 is precisely this failure shape**,
   which is why it must be fixed here rather than deferred: this PR already
   rewrites `terminal.py` substantially, so the one-line guard is inside its
   own blast radius.

### Verified sound

No OSC leak achievable — byte-by-byte splits, adjacent SGR, split escapes,
an adversarial unterminated prefix. The design also carries an undocumented
leak detector: `_at_idle_prompt` requires an exact prompt match, so any
leaked marker byte would time out every `wait_for_prompt()`. A leak cannot be
quiet.

`TIMELINE_SCHEMA` staying at 1 is correct — existing keys unchanged in name,
position, type and meaning. The smoke test is **genuinely** immune to #25
rather than dodging it: the strict checks require the literal `nonzero_exit`
and a `beat N (verb)` match, neither of which the #25 `page_error` can
satisfy.

### Non-blocking

`tests/README.md`'s new Problems/Strict prose landed after `## Known gaps`,
so an axis reads as a gap. `timeline.md` never shows `exit_code`, making the
documented `null` case invisible. `_note_issue`'s "Never raises" docstring
has no guard. Past `MAX_ISSUES` a `nonzero_exit` is counted but never
appended, so strict would not fail on it.

Sent back to the implementer.

Issue #29, opened 2026-07-25
Redaction registrations do not survive into the next segment


rogvid:
Fell out of #4.

A segmented demo builds one `Recorder` per segment:

```python
with Recorder(out, segment="part1") as rec:
    rec.redact("#api-key")
    ...
with Recorder(out, segment="part2") as rec:   # <- no redaction at all
    ...
stitch(out, ["part1", "part2"])
```

Both registries (`_secrets`, `_redacted`) are per-instance, so the second
segment records the key in the clear, and `stitch()` concatenates the two into
one demo.mp4 that is masked for its first half. Nothing warns. The same applies
to `register_secret`: a caption that fails the take in segment 1 is fine in
segment 2.

This is worse than an unredacted demo, because the first half is evidence that
redaction was considered.

## Options

- Constructor parameters (`Recorder(..., redact=[...], secrets=[...])`), so a
  storyboard can define them once and pass them to every segment. Also fixes
  the "must be called before the first goto()" ordering by construction.
- A module-level default registry every recorder picks up.
- Have `stitch()` refuse, or at least warn, when the segments it is merging
  disagree about what was redacted — it would need the timeline for each, which
  it has (`<segment>.seg.timeline.json`), and the `redact` beats are in there.

## Acceptance

A two-segment demo that registers a secret once masks it in both segments, or
says loudly that it did not.

rogvid:
Landing #7 gives this a second, worse surface. `stitch()` now merges the
segments' beat logs into one `timeline.json` / `timeline.md` next to
`demo.mp4` — and those two are **committed**, where the per-segment
`*.seg.timeline.*` were gitignored working files.

Each segment's log is scrubbed by its own recorder, against its own
`_secrets` registry. So a storyboard that registers a secret in segment two
and forgets it in segment one produces a merged, committed timeline whose
first half carries the value in plaintext — beside a second half that is
masked, which reads as evidence the log was cleaned.

Nothing in #7 changes the registry lifetime, and it should not: the fix
belongs here. Noting it so whoever takes this weighs the committed artifact,
not only the frames.

rogvid:
Rescoped by the triage in #62. The full fix — making registrations survive across segments — is not what this issue should carry; the boundary declared in #63 is that redaction covers what you name, on the recorder you name it on.

[...]

Issue #28, opened 2026-07-25
redact() is a silent no-op when its selector matches nothing


rogvid:
Fell out of #4.

`Recorder.redact("#api-ky")` — a typo, a renamed id, a selector that only
matches after a route change that never happens — registers the selector,
installs the mask, and hides nothing. Nothing warns, and the storyboard
records a take that looks redacted and is not. This is the failure mode
redaction can least afford: it presents as safe.

The recorder is already in the page on every `shot()` (`_before_shot`), so it
knows how many elements each registered selector matched.

## Proposal

Count the matches when the mask is applied, and say something when a selector
has matched nothing by the end of the take:

- a stderr warning naming the selector, at minimum;
- arguably fatal on exit, in the same spirit as `SecretLeak` failing the take
  for a captioned secret — a storyboard that asked for redaction and got none
  should not quietly produce an mp4.

Worth deciding whether "matched nothing *ever*" and "matched nothing *right
now*" are different: a selector that matches only after a click is legitimate,
so the check belongs at take level, not per-apply.

## Acceptance

A storyboard calling `redact()` with a selector that never matches anything
does not exit silently.

rogvid:
Superseded by the round-2 fix on the #4 branch: `redact()` now verifies through Playwright at every checkpoint and `_verify_redaction_final()` raises `SecretLeak` at the end of a take for any selector that never matched anything, so the take writes no mp4. The `tests/smoke` case `check_unmatched_redaction` covers it, pointing at a key inside a *closed* shadow root — unreachable by both Playwright and an injected script, which is the strongest form of this failure. Closing once #4 merges; leaving open until then so it is not closed by a branch that has not landed.

Issue #5, opened 2026-07-25
Redact secrets on the terminal PTY output path


rogvid:
Split from #1 (item 4) — the terminal half. Separate from the web redaction
slice because it is a different code path: terminal output comes from a real
shell, so there is no DOM to blur and no selector to register against.

Depends on the web slice only for the shared secret-registration API.

## Problem

`TerminalRecorder` pipes real shell output through a PTY into an in-browser
xterm. Anything the command prints is captured verbatim — API keys echoed by
a tool, tokens in a config dump, credentials in an error message.

## Design

A scrubber on the PTY→xterm write path in `_pump`, so secrets never reach the
terminal buffer that gets rendered. Two sources of patterns:

- Explicitly registered secrets (shared with the web redaction API)
- Shape-based detection: `sk-`, `ghp_`, `AKIA`, JWT structure

## Tasks

- [ ] Scrubber on the `_pump` write path
- [ ] Explicit registration, shared API with the web recorder
- [ ] Built-in shape patterns (`sk-`, `ghp_`, `AKIA`, JWT)
- [ ] Confirm the serialize addon / any screen-text capture sees the scrubbed
      buffer, not the raw one
- [ ] Document the pattern list in `SKILL.md`, and that it is best-effort —
      explicit registration is the guarantee, shape matching is the safety net

## Acceptance

A storyboard running a command that prints a registered secret, and one that
prints a `ghp_`-shaped token, produces no frame or still containing either.

rogvid:
The web half (#4) has landed, so the shared registration API this depends on
now exists on `_DemoBase` and `TerminalRecorder` inherits it:

```python
rec.register_secret(*values)   # literal text, shared registry
rec.secrets                    # -> tuple, longest first
rec.scrub(text)                # -> text with each secret -> "[redacted]"
rec._no_secrets(text, where)   # -> raises SecretLeak; for *authored* text
```

`caption()`, `interlude()` and the `.tts/` cache are already guarded for this
recorder too — they live in the base. What is left is the PTY.

**One correction to how small that looked from #4.** The web PR's report said
this would be "one `scrub()` call on the `_pump` write path". That is wrong,
and whoever picks this up should not start from it:

- **`os.read(self._fd, 65536)` returns arbitrary fragments.** A secret split
  across two reads is a substring of neither, so a per-chunk `scrub()` misses
  it — and misses it *silently*, which is the failure mode this whole slice
  exists to avoid. It needs a carry buffer holding back the last
[...]

PR #30, opened 2026-07-25
Redact secrets from frames, stills, captions, and the TTS cache (web)

### LINKED ISSUE #4: Redact secrets from frames, stills, captions, and the TTS cache (web)
Split from #1 (item 4) — the web half. The terminal PTY path is a separate
issue because it is a different code path with its own failure mode.

## Problem

No redaction exists anywhere in the skill. Demos run against
seeded-but-realistic data will leak, and a published video leaks
permanently.

## Design

Blur **in-page via CSS**, not in post. ffmpeg blur boxes need fixed
coordinates and elements move.

```python
rec.redact("#api-key", "#customer-email")       # persistent selectors
rec.type_into("#token", Secret("sk-live-..."))  # typed, never echoed
```

Implementation: an injected `MutationObserver` applying `filter: blur(8px)`
to matching selectors, registered **before the first frame** so nothing is
ever captured unblurred.

## Four leak paths, not one

- [ ] **Frames** — CSS blur via MutationObserver
- [ ] **Stills** — `shot()` must inherit the same masking. Web shots are
      currently captured full-bleed (no window frame), which makes this the
      likeliest leak path of the four
- [ ] **Captions and narration** — captions are burned in *and* spoken. Scrub
      caption text against registered secrets and **fail the take** rather
      than blur after the fact; a secret in a caption is an authoring bug
- [ ] **`.tts/` cache** — keyed by text, so a spoken secret lands on disk in
      cleartext. Scrub before caching

## Acceptance

A storyboard that renders a registered secret produces no frame, still,
caption, or TTS cache entry containing it.

## Why this is not split further

Partial redaction is worse than none: it presents as safe while leaking
through whichever path was skipped. All four web paths land together or the
slice isn't done.

rogvid:
Fixes #4.

The skill had **no redaction of any kind**. Demos run against
seeded-but-realistic data, and a published video leaks permanently. This is
the blocker on handing the skill to a team.

```python
rec.register_secret("sk-live-…")           # literal text: never captioned, spoken, or logged
rec.redact("#api-key", "#customer-email")  # CSS mask, web only
rec.type_into("#token", Secret("sk-live-…"))
```

The mask is a `<style>` rule plus inline `filter !important` plus a
MutationObserver, installed as a context init script **before the page's own
scripts**, so nothing is ever captured unblurred. Blur happens in-page, not in
post: ffmpeg blur boxes need fixed coordinates and elements move.

`Secret` is deliberately **not** a `str` subclass — `_verb_target` picks the
first string argument of a verb as that beat's `selector`, so a str-subclassed
`Secret` would write itself into `timeline.json`, a file the skill tells people
to commit.

`redact()` is deliberately **not** on the base class. A DOM selector is
meaningless for a PTY, and an inherited no-op would be a silent false promise.
`TerminalRecorder` inherits everything text-shaped (`register_secret`,
`secrets`, `scrub`, and the caption/TTS guards); #5 adds one `scrub()` call on
the PTY write path.

## How each leak path is proved, not asserted

A third smoke take graded on **pixel sharpness at native resolution** (a
downscale is itself a blur and would hand the assertion its own answer),
against an unredacted control element of identical shape **in the same frame**,
plus an absolute floor on that control — so a mask that silently fails AND a
mask that blurs everything both fail.

| | control | key masked | key unmasked | page fully blurred |
|---|---|---|---|---|
| stills | 39.3 | 0.9–1.0 | 40.6 | 0.4 |
| video | 51.7–52.2 | 1.5–3.1 | 53.6–53.8 | 0.7 |

Video is sampled at 2fps and graded on the **worst** frame — a paused video is
a screenshot. Captions/narration raise `SecretLeak` and fail the take, with the
message quoting only the scrubbed line. The `.tts/` assertion is the *exact set*
of clips, not "the leaking one is absent", which is also true of an empty
directory.

## Verification

Nine fault injections; seven caught. **Two are honestly reported as not
falsifiable in isolation** — `_before_shot` only matters once the mask is torn
out *and* the observer is gone, and the `_prepare_line` guard only matters if a
future spoken path bypasses `caption()`. Both are redundant belts, both
documented in `tests/README.md` rather than claimed as tested.

## What redaction does NOT cover

Documented at length in `SKILL.md`. The short version: a blur destroys
legibility, not information. Nothing is registered for you. Exact substrings
only — no shape detection, no unicode normalisation. `TerminalRecorder` has no
redaction yet (#5). CSS cannot reach cross-origin iframes, canvas/WebGL, or OS
dialogs. The value is still in the DOM, the network traffic, and the app's logs
— this hides it from the *recording*, not from the machine.

Filed: #28 (`redact()` is a silent no-op when its selector matches nothing),
#29 (registrations do not survive into the next segment).


rogvid:

The harness design is genuinely good — control in the same frame, absolute
floor, worst-frame grading, exact `.tts/` set comparison — and every claim
the reviewer could check was true. The mask held against mid-take `goto`, a
server-side 301, SPA `innerHTML` swaps, the tamper repair, `srcdoc` iframes,
and a canvas element. `Secret`-not-a-`str` verified: nothing anywhere
stringifies it into its value.

But the feature ships holes that make it the case we agreed is worse than
none.

### Blocking

1. **Shadow DOM: `redact()` and `type_into(Secret)` are silent no-ops.**
   ```
   page.locator('#k-open').count()             = 1
   document.querySelectorAll('#k-open').length = 0
   ```
   `click`/`spotlight`/`type_into` reach the element; the mask does not, and
   warns nothing. A live credential is typed into a clear field on camera.
   Verified by cropping the still and reading it. Lit / Stencil / LWC /
   Shoelace / Ionic — a large share of enterprise frontends.
2. **`redact()` silently ignores every non-plain-CSS selector** — `text=`,
   `xpath=`, `>> nth=` all register, install, warn nothing, hide nothing.
   It is the only verb not backed by Playwright's selector engine, and
   `SKILL.md`'s quickstart teaches `rec.click("text=Orders")`.
3. **The pixel assertion cannot measure legibility.** Blur is a constant
   8px that does not scale with text size, and the control's edge energy
   *falls* as the font grows while the blurred value stays flat. At 52px the
   secret is transcribable off a frame the harness passes; at 68px the
   control approaches the anti-vacuity floor for the wrong reason. Injecting
   `REDACT_BLUR_PX = 3` leaves every pixel assertion green.
4. **Two unguarded paths:** `terminal_close(stamp)` renders a registered
   secret large in frame; the beat log scrubs `selector` but not `still` or
   `caption`, so `"selector": "04-[redacted]"` sits beside
   `"still": "images/04-sk-live-….png"` in a file the skill says to commit.

### Correction to the PR's own tally

Three guard sites are individually untested, not two. Removing `caption()`'s
`_no_secrets` leaves `--redaction-only` passing, because `_prepare_line`'s
guard covers the same path — the relationship is symmetric, not
belt-and-braces.

### The safety contract needs the omissions

Shadow DOM; CSS-only selectors; the non-scaling blur radius;
`terminal_close()`; already-attached iframes; the `still`/`caption` fields.
Also **the canvas claim is wrong in the safe direction** — `filter` on a
`<canvas>` does blur its rendered pixels; only bitmap readback is
unaffected. As written it talks people out of a working mitigation.

### #5's scope needs amending

"One `scrub()` call" does not hold: `os.read()` returns arbitrary PTY
fragments, so a secret split across two reads is a substring of neither.
Needs a carry buffer. ANSI escapes can interleave inside a token, and
`run()`/`send()` echo authored text with no guard.

Sent back to the implementer.

rogvid:
## Fourth review: BLOCK

**Confirmed closed:** L2 (dialect refusal, no bypass found), R1 (navigation
blanking gone — `longest_blank=0.00s` across click/back/submit), R2
(cross-frame masking, and it reaches cross-origin and opaque-origin frames,
which SKILL.md says it cannot). The rebase preserved upstream strict mode:
`artifacts kept` on `StrictTakeFailed`, nothing kept on `SecretLeak`, not
conflated.

### The structural finding

**`applied < needed` is unreachable by construction.**
`__demoRedactMark(el)` calls `wear(el)`, which sets `filter` to exactly
`needed(el)` at `!important`, then returns `applied(el)` read from that same
computed style. It compares a number against itself. The whole safety
property is `needed()`; verification adds no independent evidence — which is
why the five holes below pass silently instead of failing loudly.

### Five new leaks, all in `textPx()`

All reported `applied=8.00 needed=8.00 text=15.00` over 60-80px glyphs, all
readable in a still and an mp4 frame of a take that reported success:

| | shape | why missed |
|---|---|---|
| A1 | `::after` at 64px | `getComputedStyle` never called with a pseudo argument |
| A2 | descendant `transform: scale(5)` | computed font-size is pre-transform |
| A3 | descendant `zoom: 5` | computed 13px, rect height 72 |
| A4 | SVG `<text>` scaled by viewBox | `REPLACED` is uppercase; SVG `tagName` is lowercase |
| A5 | shadow root **one level below** the target | only `el.shadowRoot` is consulted |

**A5 is blocking on its own** — SKILL.md documents redacting a wrapper *and*
reaching open shadow roots; a Lit/Shoelace card is the composition of the
two.

### Direction

Stop deriving text size from CSS — it has unbounded ways to scale rendered
text and we have found five. Derive the radius from rendered geometry
(`getBoundingClientRect`, which includes transforms and zoom), recurse shadow
roots at every depth, and pass pseudo-elements to `getComputedStyle`. The
replaced-element box-height proxy is the one part that could not be broken;
generalise it. If geometry cannot close it, switch the default to opaque
erasure — a blur is inherently a "how much is enough" question, an opaque box
is not.

Marked as the final fix round; after this it ships with a narrowed contract
if needed.

Issue #26, opened 2026-07-25
TerminalRecorder's PTY child does not inherit the determinism controls


rogvid:
Split out of #10, which landed the browser-side determinism controls (frozen
wall clock, pinned timezone/locale, motion disabled).

Those controls are Playwright *context* options and page init scripts. The
`TerminalRecorder`'s program does not run in the page — it runs under a PTY as
a child process, and its environment comes from `os.environ.copy()` in
`TerminalRecorder._start()`. So:

```
❯ date
Sat Jul 25 18:20:51 CEST 2026     # the machine's real clock, in its own zone
```

A terminal demo that shows a timestamp — `date`, `ls -l`, `git log`, a log
line, a build duration — therefore still records differently every take, while
the web recorder next to it records identically. #10's acceptance criterion is
met for the browser and silently not met for the terminal.

`tests/README.md` lists this under "Known gaps"; `SKILL.md` says plainly that
the terminal child is outside the frozen clock.

## Directions, none tried

- **`TZ` only.** `env["TZ"] = self._timezone_id` when `deterministic` pins the
  *zone*, so at least the formatting is stable. Cheap, and half a fix: the
  instant still moves.
- **`faketime`.** `libfaketime`/`datefake` in front of the shell would freeze
  the instant too, but it is an external dependency this skill does not have
  and cannot assume, and it changes what the demoed program observes in ways
  the storyboard author did not ask for.
- **Say no, loudly.** Document it as out of scope and leave the storyboard
  author to pin their own program's clock. Cheapest, and possibly right — the
  recorder cannot seed an app's RNG either.

Whichever it is, it changes what a user's terminal demo records, so it wants
its own review rather than riding along with #10.

## Acceptance

Either a terminal demo that prints a timestamp reproduces across takes, or
`SKILL.md` states the limitation in the same place it states the RNG one and
this issue closes as won't-fix with that reasoning recorded.

rogvid:
Closing into #63 as a stated limit — this is already stated in SKILL.md and in tests/README.md's Known gaps; #63 consolidates it into one place. A terminal demo showing `date`, `ls -l` or `git log` records differently every take. Worth saying: the TZ-only half-fix (`env["TZ"]` when `deterministic`) remains cheap, and if a pilot demo trips over this, reopen with that scope rather than the full one. Scope record: #62.

Issue #18, opened 2026-07-25
Playwright's screencast loses ~0.6s of wall time mid-take, so late beat timestamps drift from demo.mp4


rogvid:
Found while verifying the acceptance criterion of #6 ("`timeline.json` beat
timestamps land within ~100ms of the corresponding caption appearing in
`demo.mp4`"). The beat log is on the right clock; the *video* is not.

## What was measured

A storyboard flashed the caption on and off at known `time.monotonic()`
offsets from the start of capture, and the recording was searched for each
flash (caption-band luma, 25 fps, so ±40 ms). One 32 s take:

```
  wall   1.979  video   1.880  offset -0.099
  wall   4.991  video   4.960  offset -0.031
  wall   7.997  video   7.960  offset -0.037
  wall  11.015  video  11.000  offset -0.015
  wall  14.024  video  14.000  offset -0.024
  wall  17.042  video  16.440  offset -0.602   <-- step
  wall  20.059  video  19.480  offset -0.579
  wall  23.072  video  22.480  offset -0.592
  wall  26.084  video  25.520  offset -0.564
  wall  29.088  video  28.520  offset -0.568
```

A single discrete **~0.57 s of wall time that never reaches the webm**, after
which every frame sits that much earlier in the video than the clock says.
It is not a rate mismatch — the offset is flat either side of the step.

Across ten `tests/smoke` takes, five stalled once (closing caption -520 to
-680 ms) and five did not (-40 to -120 ms). The take's first caption, ~2 s
in, was -40 to -120 ms in **every** take. `demo.mp4`'s duration shrinks by
exactly the lost amount, which is how the stall can be detected after the
fact even though the recorder cannot see it happen.

An isolation run points at idle stretches: a take that only screenshots lost
nothing, a take that idles for 10 s lost ~1.0 s, a take that kept the page
changing for 10 s lost ~0.5 s. Chromium's screencast emits frames on change,
and nothing pads the gap when there is none.

## Why it matters

- #8 extracts review frames *at* beat timestamps. In a stalled take the frame
  a reviewer is handed can be more than half a second off the caption it is
  labelled with — which is precisely the reviewer-hallucination failure #8
  exists to remove.
- #9's per-beat evidence inherits the same offsets.
- #7 merges segment timelines by real segment durations, so each segment
  carries its own stall into the merged file.

`tests/smoke` currently grades this with two bars, documented in
`tests/README.md` under "Known gaps": 200 ms on the take's first caption (the
[...]

PR #31, opened 2026-07-25
Determinism controls: frozen clock, fixed locale, disabled animation

### LINKED ISSUE #10: Determinism controls: frozen clock, fixed locale, disabled animation
Split from #1 (item 6). Independent — can land any time.

## Problem

Re-recording produces a different video every time, so visual diffing is
impossible and "did the UI actually change?" is unanswerable. Timestamps,
locale-formatted values, and animation phase all vary between takes.

## Tasks

- [ ] Freeze `Date.now()` / `new Date()` via an init script on the context
- [ ] Fixed timezone and locale on the browser context
- [ ] `prefers-reduced-motion: reduce` plus injected CSS disabling animations
      and transitions
- [ ] Document RNG/data seeding expectations in `SKILL.md` — the recorder can
      only control the browser, the app's own randomness is the storyboard
      author's problem and should be called out as such
- [ ] `Recorder(..., deterministic=False)` escape hatch for demos where the
      frozen clock breaks the app

## Acceptance

Two consecutive runs of the same storyboard against unchanged app state
produce videos that differ only within encoder noise.

rogvid:
Fixes #10.

Re-recording produced a different video every time, so visual diffing was
impossible and "did the UI actually change?" was unanswerable.

Adds a frozen wall clock, pinned timezone/locale, `prefers-reduced-motion:
reduce` plus a motion-flattening CSS rule, and a
`Recorder(..., deterministic=False)` escape hatch. **Default is on**, which is
the behaviour change the issue asks for.

## The freeze is of the wall clock only, deliberately

`Date.now()` and zero-arg `new Date()` answer from a fixed epoch via a `Proxy`
on `Date`. `performance.now()`, `requestAnimationFrame` and `document.timeline`
are untouched — because freezing monotonic time would stop the compositor,
which is the exact mechanism of #18. It would *cause* the stall rather than
reproduce a video.

The motion rule carries a published opt-out, `data-demo-video-animate`, for
elements that must keep the compositor awake. Motion is flattened via duration
0 + `iteration-count: 1` + `fill-mode: forwards` rather than `animation: none`,
so apps that animate content in from `opacity: 0` land on the finished state
instead of staying permanently invisible.

## What "deterministic" is asserted to mean

One storyboard recorded three times against a new `?entropy=1` fixture hook:

- **Stills: byte-identical `sha256`** across two deterministic takes. Lossless
  PNGs, so no threshold and no encoder noise — sensitive enough to catch a
  four-digit timestamp change.
- **The same stills inside the `deterministic=False` take must differ**, which
  is what stops "two takes match" from being the vacuous "two blank pages are
  equal".
- **Video, honestly bounded:** H.264 is not byte-reproducible, so closing
  frames are compared over the entropy panel. Deterministic pair 0.00-0.51, vs
  off 4.16-4.93, bars at 1.5/2.5 in the gap, both directions asserted.
  `tests/README.md` states what this cannot see: it caught a whole panel
  changing colour (24.14) and did *not* notice four digits changing (0.22).
- **Page-reported state** read from inside the page, never from
  `self.deterministic`, and inverted in the escape-hatch take.

The ticker guard plants a **control element** carrying the ticker's animation
with no opt-out: the take fails unless the rule flattened the control *and*
left the ticker running. Checking only the ticker would pass equally on a
recorder that had stopped injecting the rule at all.

## Verification

8 consecutive full runs with determinism on, all passed, worst skew −240ms with
both probes moving together — nowhere near the −500…−700ms stall signature. The
injection that removes the ticker's opt-out reproduced the stall on the first
try at −640ms, so the opt-out is demonstrably what holds it off.

13 fault injections, each watched fail. Two path-existence checks were not
separately injected (they duplicate `check_take()`), and one assertion was
*deleted* because it could not be made to fail — noted in a code comment.

Honest finding: headless Chromium reproduces animation phase across
identically-paced takes well enough that a spinner exempted from the motion
rule produced byte-identical stills in both takes, and was caught only by the
within-take comparison. Documented rather than papered over.

Filed #26 — `TerminalRecorder`'s PTY child does not inherit the determinism
controls.

### DIFF
--- a/skills/demo-video/README.md
+++ b/skills/demo-video/README.md
@@ -31,6 +31,13 @@ and a real terminal session (voice on):
 - **`record.py`** — the storyboard that produced the media, and the thing
   that actually gets committed: re-runnable after the UI changes, so the
   video stays out of git history and is regenerated rather than archived.
+- **A recording that reproduces, when you ask for it** — the browser's
+  timezone and locale are always pinned, and `Recorder(deterministic=True)`
+  additionally freezes the page's clock and flattens animations, so the same
+  storyboard gives you the same stills rather than a new set of timestamps.
+  Opt-in because a stopped clock changes what a debounce, a token check or an
+  elapsed-time bar does, usually without saying so; `SKILL.md` shows the five
+  shapes that break and what the recorder cannot pin at all.
 - **Spoken narration (optional)** — with `ELEVENLABS_API_KEY` set, every
   caption line is synthesized via ElevenLabs and mixed onto the mp4 at the
   moment it appears. Clips are cached; pacing self-adjusts so speech is
@@ -80,6 +87,10 @@ win over env vars.
 | `DEMO_VIDEO_TERMINAL_SHELL` | shell `TerminalRecorder` launches | `/bin/bash` |
 | `DEMO_VIDEO_TERMINAL_FONT_SIZE` | `TerminalRecorder` font px | `15` |
 | `DEMO_VIDEO_VIEWPORT` | recording size, `"1280x720"` | 1280×720 |
+| `DEMO_VIDEO_DETERMINISTIC` | freeze the page clock and flatten motion (`1`/`0`) | **off** |
+| `DEMO_VIDEO_CLOCK` | instant the clock is frozen at, when it is (ISO 8601) | `2025-01-01T09:00:00Z` |
+| `DEMO_VIDEO_TIMEZONE` | browser timezone (always applied) | `UTC` |
+| `DEMO_VIDEO_LOCALE` | browser locale (always applied) | `en-US` |
 | `DEMO_VIDEO_SPEECH` | force narration on/off (`1`/`0`) | auto by API key |
 | `DEMO_VIDEO_VOICE_ID` | ElevenLabs voice | Sarah (premade) |
 | `DEMO_VIDEO_SPEECH_MODEL` | ElevenLabs model | `eleven_multilingual_v2` |
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -122,6 +122,10 @@ clean:
 | `DEMO_VIDEO_TERMINAL_SHELL` | shell `TerminalRecorder` launches | `/bin/bash` |
 | `DEMO_VIDEO_TERMINAL_FONT_SIZE` | `TerminalRecorder` font px | `15` |
 | `DEMO_VIDEO_VIEWPORT` | recording size, `"1280x720"` | 1280×720 |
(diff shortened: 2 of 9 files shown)


rogvid:
## Cold review: BLOCK — on the default, not the feature

Fault-injection discipline is the best of the three PRs in flight. The
byte-identity assertion is real (injections 1 and 3 produced genuine digest
mismatches), the `deterministic=False` inverse fails when it should, and
6/6 independent full runs confirmed the #18 flake did not return — worst
skew −240ms with both probes moving together. Selector performance is a
non-issue (44.7ms vs 44.3ms on 20,009 nodes).

### Blocking

1. **Default-on silently breaks five of five ordinary app shapes.** A
   lodash-shaped debounce never fires (`time - lastCallTime` is always 0, so
   the timer reschedules forever); a progress bar sticks at `0%`; a JWT gate
   renders "Session not yet valid (clock skew)"; a "last 7 days" chart
   renders zero bars; a `transitionend` drawer never opens. **Four of five
   fail silently** — no exception, no warning, nothing in `timeline.json`,
   just a plausible wrong screen. The one loud failure wedges 156s then
   `TimeoutError` with **no mp4 written** and nothing naming the clock.
   → **Decision: clock freeze becomes opt-in.** Timezone/locale pinning and
   `prefers-reduced-motion` stay on by default.
2. **No discoverability.** Four `print()` sites in `core.py`, none about
   determinism; the narration notice is a whole paragraph. `timeline.json`
   records no determinism settings, so a committed still carries no record
   of which clock produced it — the exact question the feature exists to
   answer.
3. **The motion rule kills `transitionend`** (duration 0 never starts a
   transition, so the event never fires — accordions, modals and wizards
   stall), and **"lands on the finished state" is false** for `alternate`,
   `alternate infinite`, and delayed fade-outs. A snackbar with
   `animation-delay: 4s` is **invisible from frame zero**.
4. **The freeze is incomplete, so stills are *almost* reproducible.** Two
   fully-deterministic takes of one storyboard produced 2/2 differing
   stills. `Intl.DateTimeFormat().format()`, Workers,
   `new Date().constructor.now()`, `performance.timeOrigin` and
   `document.lastModified` all read live time. Also
   `Date.prototype.constructor === Date` is now `false` — a correctness
   break independent of time.
5. **One injection was missed:** dropping `::before`/`::after` coverage
   passes clean. An animated `::after` is the most common CSS spinner on the
   web.

### Verified correct

`instanceof`, `getPrototypeOf`, subclassing, `Reflect.construct`, `Date()`
as a function, `JSON.stringify`, `structuredClone`, `Date.parse`,
`Date.UTC`, explicit-argument constructors, and both `about:blank` and
`srcdoc` iframes.

Sent back to the implementer.

Issue #50, opened 2026-07-25
Evidence records url and title, which redact() cannot reach


rogvid:
Found in review of #9.

Each beat's `evidence/beat-NN.json` records `url` and `title`. Both go
through the registered-secret scrub, and nothing else touches them.

That is a real hole, because the two mechanisms that make the rest of an
evidence file safe do not apply to either:

- `redact()` covers where a value **renders**. A token in a query string is
  not rendered by anything, so no selector names it and the harvest that
  turns redaction into text masking never sees it.
- The value is not a caption, so `register_secret()`'s "fail the take" path
  is not involved. If the author did not register it, it is written out.

A magic-link URL, a `?token=`, an `access_token` in a fragment, a session id
in a path segment — all ordinary things for a demo to navigate through, and
all of them land in a plaintext file, once per beat, for the whole take.

`redact()` cannot help here even in principle. The honest options:

- scrub the query string and fragment down to parameter *names* by default
  (`?token=[redacted]&page=2`), with a way to keep values that matter to the
  story;
- allow `redact_url("token", "code")` naming parameters to mask;
- or drop `url` to origin + path and put the rest behind an opt-in.

The first is probably right: a reviewer reading evidence wants to know *which
page* the demo was on, and almost never wants the query values.

`title` is the same argument with a smaller surface — an app that renders a
customer name into `<title>` is common, and it is not covered by a redact()
selector either.

## Acceptance

A take that navigates to a URL carrying an unregistered secret, and an
assertion that the secret is in no evidence file — with the *path* still
readable, so the fix is not "stop recording the URL".

rogvid:
**blocks-v1** (#62), and sharper than when it was filed. `evidence/` is neither in SKILL.md's committed-files table nor in `.gitignore`, so on current defaults a `?token=` in `page.url` lands in **git history**, once per beat, permanently. Neither mechanism protects it: `redact()` covers where a value renders and a query string renders nowhere; `register_secret()` only fires on captions. Decide what `evidence/` is (committed artifact or working file) as part of this.

rogvid:
Rescoped by the delivery decision in #2. Evidence is now a **CI step artifact only** — read by the reviewing agent inside the same pipeline, never published, never committed.

[...]

Issue #49, opened 2026-07-25
Per-beat evidence: the unconditional whole-page ARIA snapshot is the cost, not the harvest


rogvid:
Found in review of #9. The same shape as #44, one layer up.

Per-beat evidence costs, for **every** beat of a web take:

- a `_checkpoint()` (which is itself N round trips per redacted selector, per
  #44),
- one `frame.evaluate` per frame to harvest what `redact()` covers,
- `page.url`, `page.title()`,
- `locator("body").aria_snapshot()`,
- and, when a spotlight is up, a second `aria_snapshot` plus one
  `evaluate` for the scoped `outerHTML`.

## What is known

On `tests/fixture` — a 24 kB static page with two frames — measured at the
Playwright API rather than in a take: `aria_snapshot(body)` 24 ms,
`outerHTML` 4 ms. The 23-beat web take grew from 20.1 s to ~20.7 s, i.e.
about 25 ms of added wall time per beat, and the 31-beat redaction take (nine
redacted selectors, two frames) from 20.9 s to ~22 s.

## What is not

Every one of those numbers is from a page with 60 elements. `aria_snapshot`
serializes the whole accessibility tree, so its cost scales with the
document, and a real app's tree is not 2.3 kB — the cap exists because it can
be hundreds of kilobytes. Nobody has recorded a real application with this
on and looked at what it did to pacing.

That matters more here than raw overhead: the time is spent **inside** the
beat, deliberately (so no gap opens between beats), which means it lands
between the verb returning and the next one starting — i.e. inside what a
viewer sees as the pause after an action.

## Options if it turns out to be too much

- capture only at `caption`/`spotlight`/`shot` beats, which is what #9's
  proposal literally said before its acceptance criterion widened it to every
  beat;
- reuse one harvest across a short window instead of re-running it per beat;
- make the ARIA snapshot the only thing captured for cheap verbs.

## Acceptance

A measurement on a real (non-fixture) app of the added wall time per beat and
the size of `evidence/`, and either a note in `SKILL.md` saying it is
negligible or one of the options above.

rogvid:
**Correcting this issue: it is scoped to the wrong half.** Measured under
adversarial review of #9, and the numbers move the target completely.

## The harvest is not the cost

`_harvest()` — the page evaluation that reads what `redact()` covers, plus
what the page renders outside it — is **~13 ms at 20 000 nodes**, and close to
flat in the number of redacted selectors, because it resolves them all in one
[...]

Issue #48, opened 2026-07-25
Evidence: the off switch and the refuse-to-write path are both unexercised


rogvid:
Found in review of #9.

Per-beat evidence has two paths that exist to keep a take safe or cheap, and
`tests/smoke` exercises neither.

## `evidence=False` / `DEMO_VIDEO_EVIDENCE=0`

Every take in the harness records with evidence on. The off switch turns off
the `evidence` key on beats, the capture at every beat close, and the
`evidence/` directory — three behaviours, none asserted. A regression that
made the flag do nothing would pass the whole suite, and so would one that
made it crash: nothing calls it.

Worth having because the flag is not decoration. Evidence costs a handful of
Playwright round trips per beat, and a project recording a 200-beat demo
against a slow app is the case it is there for.

## The `omitted` path

When `redact()` is registered and the recorder cannot read what the mask is
covering, `Recorder._evidence_payload` returns
`{"omitted": "..."}` and the beat's evidence is written with **no page text**
— the deliberate refusal, because a DOM dump is not protected by a pixel
cover and a mask the recorder cannot read is a mask it cannot mask against.

Provoking it needs a main frame that fails mid-capture, which no fixture here
manufactures. So the branch that refuses rather than guessing is read-only
coverage today, and it is exactly the branch a reviewer would most want
proven.

## Acceptance

- One take (or one assertion on an existing short take) records with
  `evidence=False` and requires: no `evidence/` directory, and no `evidence`
  key on any beat in `timeline.json`.
- The `omitted` path is provoked — a monkeypatched harvest that raises is
  the cheap version — and asserted to leave a file that exists, names the
  beat, carries `omitted`, and carries no `aria`, `html` or `screen`.

rogvid:
Closing as part of a backlog cull, not because the finding is wrong.

Every issue in this batch is accurate and most carry measurements. They are
being closed because **the tracker had become a record of everything known to be
imperfect rather than a list of work anyone intends to do** — 58 open against a
`blocks-v1` list of 8. An issue nobody will schedule is worse than no issue: it
makes `gh issue list` useless, which is the command CLAUDE.md says to start every
session with.

The bar applied: *would this be scheduled?* Not *is it real?*

This one falls into one of three groups:

- **A test could grade more than it does**, with no known defect behind it. Worth
  doing if someone is already in that file; not worth its own slot.
- **A micro-nit** — a field name, a message, a small cost — with no user-visible
  consequence.
[...]

Issue #47, opened 2026-07-25
tests/README.md states several of its own limits twice, in contradictory versions


rogvid:
Found while adding the evidence axis in #9.

`tests/README.md` carries several paragraphs twice, in versions that
contradict each other. They look like the residue of two PRs editing the
same section and both being kept.

## What is duplicated

- **The axis count.** "Six independent axes, because a recorder can fail on
  any one of them while looking perfect on the other five." is immediately
  followed by "Five independent axes, because a recorder can fail on any one
  of them while looking perfect on the other four."
- **"Re-running into the same `--out-dir` is safe"** appears twice, once
  listing the ten take subdirectories and once listing only three
  (`web/`, `redaction/`, `terminal/`).
- **"Deleting is bounded"** appears twice, with the same disagreement about
  which directories are ever removed.
- **"Nothing checks audio"** appears twice under Known gaps: once saying the
  whole speech path is untested, once saying the redaction take exercises it
  against a stub. The second is the true one; the first was written before
  the redaction take existed.
- **The fixture's query-string hook table** appears twice, with different
  rows — one describing `?secret=1` as rendering a single `#api-key`, the
  other as the whole panel.

## Why it matters

This file is the one place the harness's *limits* are written down, and
Known gaps is explicitly there so that "an assertion nobody knows is missing"
does not happen. A section that states a gap and then states its opposite
forty lines later is worse than either, because a reader who stops at the
first one is misinformed and cannot tell.

## Deliberately not fixed in #9

Reconciling them means rewriting sections that #5, #7 and #8 are all queued
to edit. It is a doc-only change and wants its own commit, once that queue
drains — the same reasoning as #13.

## Acceptance

`tests/README.md` states each of the above once, in the version that matches
what `tests/smoke` actually does, and the axis count matches the number of
axes described.

rogvid:
Folding in #55, which is the same bug found independently while adding the
terminal-redaction section for #5. It carries line references worth keeping:

- `## Running it` code block lists `--web-only`/`--terminal-only` twice, once as
  "all three takes" and once as "every take" (~L27-37)
- `Re-running into the same --out-dir is safe:` twice, once naming ten take
  directories and once naming three (~L78-94)
- `Deleting is bounded.` twice, same split (~L96-103)
- `## What it asserts` opens with both "Six independent axes … other five" and
  "Five independent axes … other four" (~L127-130)
- the fixture hook table lists `?console-error=1` and `?secret=1` twice with
  different descriptions
[...]

PR #52, opened 2026-07-25
Capture agent-readable evidence per beat

### LINKED ISSUE #9: Capture agent-readable evidence per beat (a11y snapshot, outerHTML, screen text)
Split from #1 (item 2). Depends on the beat event log slice.

## Problem

A reviewing agent infers DOM state from pixels. The recorder is driving the
page — it has the real thing. That round trip through images is lossy and
expensive, and it is unnecessary.

## Proposal

At each caption/spotlight, capture alongside the frame:

- `page.accessibility.snapshot()` scoped to the spotlight target — semantic,
  compact, and stable across restyling; preferred over raw HTML
- `outerHTML` of the spotlight target
- For `TerminalRecorder`: the ANSI-stripped rendered screen (`_screen()`
  already returns this)

Store as `evidence/beat-NN.json`.

## Tasks

- [ ] Web: a11y snapshot + outerHTML per beat
- [ ] Terminal: screen text per beat
- [ ] Cap snapshot size; truncate with an explicit marker rather than bloating
      the evidence folder
- [ ] Ensure evidence inherits redaction if that slice has landed — this is a
      fifth leak path and it is plain text

## Acceptance

`evidence/beat-NN.json` exists for every beat, is size-capped, and a reviewer
given only the evidence files can state what was on screen without seeing an
image.

rogvid:
Fixes #9.

A reviewing agent infers DOM state from pixels. The recorder is driving the
page — it has the real thing, and the round trip through images is lossy and
expensive.

Writes `evidence/beat-NN.json` per beat: an ARIA snapshot, the spotlight
target's `outerHTML`, and for `TerminalRecorder` the ANSI-stripped screen.

## Spec correction

The issue names `page.accessibility.snapshot()`. **That API no longer exists**
— removed from Playwright (verified: `Accessibility` is not importable from
`playwright.sync_api` at 1.61). Uses `locator.aria_snapshot()`, its successor,
which is what the issue actually asked for ("semantic, compact, stable across
restyling"). No fallback branch — an older Playwright gets `aria: null` and an
`aria_format` saying why, rather than a second code path nothing can exercise.
`tests/smoke` pins `playwright>=1.49`.

`outerHTML` is only ever the spotlight target's, never the page's: body markup
measured 24 kB against 2.3 kB of ARIA, and carried inline `<script>` text and
`srcdoc` attributes — source code and whole embedded documents nobody put on
screen.

## Evidence is a fifth leak path, and it needed its own mechanism

`redact()` covers *pixels*. The value stays in the DOM. Evidence dumps the DOM.
`register_secret()` does not help, because `redact()` never tells the recorder
what the element says.

So `_redacted_rendered_text()` reads it out of the page at every capture —
everything matching a registered selector or the mask's marker, walking every
node in the subtree, shadow roots at every depth, `::before`/`::after` computed
content, input values and value-bearing attributes, across every frame. That
union is applied to all beats on the way out, so a value first seen at beat 20
is masked out of beat 3.

Harvesting only the element the selector matched **left 3 of the fixture's 8
keys in the clear** — a value in a redacted *wrapper*, one in a `::after`, and
one two shadow roots down. Markup also needs structural elision:
`sk-live-<b>FAKE</b>` has a `textContent` a string mask finds and an
`outerHTML` it does not.

Masking runs **before** capping, deliberately: capping first can cut a secret
in half and leave a prefix no substring search finds.

## Proof

The byte sweep requires all 13 secrets absent **and both unredacted controls
present** — otherwise "no secret appears" is a statement about an empty
directory:

```
smoke: redaction none of the 13 secrets appears verbatim in any of the 45 files
the take wrote (31 of them per-beat evidence, which the controls prove is not empty)
```

With the harvest disabled, exactly the seven values that only `redact()` covers
leak into 26 files each while the two *registered* ones stay masked — the
asymmetry, demonstrated.

Twelve fault injections, each watched fail.

## Not committed

`evidence/` is gitignored: greppable plaintext of a real app's DOM, regenerated
wholesale per take, and git cannot be made to forget it. `timeline.json`/`.md`
remain the committed record.

## #22

Settled at no cost to #7: the segment goes in the filename and the path is
written **onto the beat** as `evidence` rather than derived from `index`.
SKILL.md says "read the pointer, never rebuild it". #7 renumbers freely.

Filed: #47, #48, #49, #50.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ frames/
 # Left behind only when a take crashes before the recorder cleans up.
 .video/
 .frame.png
+# Per-beat evidence: greppable plaintext of the app's DOM (or terminal),
+# regenerated wholesale on every take. SKILL.md says why it is not committed.
+evidence/
 
 # Smoke-test recordings. tests/smoke records into a temp dir by default; this
 # covers pointing it back into the repo (tests/smoke --out-dir tests/out).
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,8 @@ frames/
 .frame.png
 # Per-beat evidence: greppable plaintext of the app's DOM (or terminal),
 # regenerated wholesale on every take. SKILL.md says why it is not committed.
+# It is read by the reviewing agent in the same pipeline that produced it and
+# nowhere else (issue #50).
 evidence/
 
 # Smoke-test recordings. tests/smoke records into a temp dir by default; this
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -58,6 +58,36 @@ The words, and what each fixture element they belong to is for:
   SDCT  #sd-token-control  the same, inside the shadow root
   LATE  a caption          set before the value in it was registered
 
+...and the ?evidence=1 panel (issue #9), whose shapes leak into a *text* dump
+of the page rather than into a picture of it:
+
+  WSPC  #ev-ws-key         indented on its own line in hand-written markup
+  BLED  #ev-bleed-key      in a card whose label also renders outside it
+  CHLD  #ev-child-key      split into text nodes that are ordinary on their own
+  LBLD  #ev-label-src      an accessible name sourced from outside the subtree
+  SLOT  #ev-slot-light     light DOM slotted into a redacted shadow host
+  VALU  #ev-value-key      an input value that lives on the property
+  TICK  #ev-tick-key       rewritten every 25 ms with a value nothing has seen
+  ATTR  #ev-spot-attrs     a data-* attribute the page never renders
+  JSON  #ev-spot-attrs     the same, inside a JSON blob in another attribute
+
+...and three that never render at all — they exist only inside tests/smoke, one
+per assertion about how the writer handles a value it was told about:
+
+  PRNT  a printed line      echoed by a command in the terminal take, where
+                            register_secret() is the only defence (issue #5)
+  STAL  a planted file      a previous take's evidence, which a re-record into
+                            the same folder has to delete
+  CUTT  a padded list item  positioned so the markup cap cuts through it, which
+                            is the only place "mask, then cap" is observable.
+                            It is the one with more than sixteen zeroes — that
+                            length is what makes the straddle arithmetic land,
+                            hence the wider run below
+
+The word list is what keeps this narrow, not the run length: every literal is a
+four-letter word followed by nothing but zeroes, and lengthening the run admits
+no string the word list did not already admit.
+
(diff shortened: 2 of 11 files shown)


rogvid:

**The byte sweep does not grade anything.** Replacing `_evidence_payload`
with a constant returning only the two control strings passes
`--redaction-only`, printing verbatim the sentence this PR quotes as its
proof. And `html` is `null` in all 31 swept files — the redaction
storyboard never calls `spotlight()`, so the entire `outerHTML` surface is
never swept. The take with secrets is not graded on evidence content; the
take graded on evidence content has no secrets.

**Six leaks, four on values `redact()` covers in pixels:**

- **L1 whitespace** — a value on its own line in hand-written HTML. Harvest
  gets `'\n  wombatxray7714\n'`, ARIA gets `wombatxray7714`, `str.replace`
  finds nothing. **12 of 13 files.** The fixture escapes it only by accident
  (every key is set via `el.textContent` in JS).
- **L3 TOCTOU** — harvest and snapshot are separate round trips; a 25ms
  ticker leaks the next value. Any countdown or re-rendering token field.
- **L5** — a `register_secret()` value split by inline markup outside a
  redacted element; `hide()` fires only on marked elements.
- **L6** — `outerHTML` keeps `data-*`, `title`, `alt` on non-redacted nodes.
  **None of these ever rendered.** A new exposure surface, kept on the same
  rationale used to strip `<script>` and `srcdoc`.
- L2 (`aria-labelledby` from outside the subtree) and L4 (slotted light DOM)
  are defensible as documented limitations.

**A fail-open regression:** `_capture_evidence`'s bare `except Exception`
swallows `SecretLeak`, and `_evidence_payload` adds a new `_checkpoint()`
call inside a `finally`. Four leaks downgraded to a warning; mp4 and
timeline written. Before this PR every `_checkpoint()` failure killed the
take.

**Mask bleed, the mirror defect:** redacting a card containing
`<span>Revenue</span>` registers "Revenue" as a global forbidden literal.
One run registered `"sk-"` and `"live"` and rewrote every unrelated key in
every file.

**Stale evidence is never cleared** — re-recording into the same folder
leaves earlier beats holding now-redacted values.

**4 of 6 re-run injections missed.** The `<segment>.seg.beat-NN.json`
naming has never run, so "#22 settled" is untested.

Verified correct: the Playwright pin (`>=1.49` is exactly right,
`aria_snapshot` is genuinely the successor), and no leak via iframes,
canvas, `::after`, or split markup inside a redacted subtree.

Issue #51, opened 2026-07-25
stitch()'s issue-merging path is untested: the segmented take is healthy by design


rogvid:
Found while landing #7.

`stitch()` merges each segment's `issues` as well as its beats: every issue's
`t` is offset by the cumulative duration of the preceding segments, and its
`beat` is re-pointed from the segment's own beat list at the merged one
(`base + beat`, where `base` is how many beats have already been merged).

Neither line is exercised. `tests/smoke`'s `segments/` take records the healthy
fixture under `strict=True` — deliberately, because that is the only assertion
that can fail on *over*-reporting — so it records no issues at all, and every
statement above is covered by nothing.

What that leaves open: an issue from segment two arriving in the merged
timeline attributed to whatever beat of segment one happens to sit at that
index. That is a *confidently wrong* attribution, which the beat-attribution
work in #3 argues is worse than `beat: null` — and it lands in the one file a
reviewer reads to find out what broke.

## Options

- A short third segment recorded against `?console-error=1`, graded by
  `check_issues()` on the merged timeline, with the expectation naming the
  segment as well as the beat. Costs one more browser launch (~8 s).
- Or grade the existing `web-problems/` take as two segments instead of one,
  which adds no take but changes what that take is for.

## Acceptance

A merged timeline whose second segment recorded a problem attributes it to a
beat of *that* segment, by verb and by segment name, and its `t` is on the
stitched video's clock.

rogvid:
**blocks-v1** (#62), category: the artifact must never lie. `base + beat` re-pointing is unexercised because the `segments/` take is healthy by design. The failure mode is an issue from segment two attributed to whichever beat of segment one sits at that index — confidently wrong attribution, which #3 already argued is worse than `beat: null`.

rogvid:
**Declared rather than fixed**, and now documented in [`skills/demo-video/reference/limits.md`](https://github.com/rogvid/skills/blob/main/skills/demo-video/reference/limits.md) under *what a stitched demo's artifacts promise* (PR #167), with the measurement from this issue carried across so the evidence survives the close.

This is the `verified-review` rule for asymptotic properties: a reviewer will always find another hole in "notice everything about the app" or "reproduce every take", so a tracker that records every hole grows without bound and stops being a plan. The boundary is the deliverable — an author who knows it can work around it.

**Every premise was re-verified against the code before being written up**, not copied from the issue text. Reopen if you want it *fixed* rather than declared; the measurement is in `limits.md` either way.

Issue #22, opened 2026-07-25
Beat index is positional per segment, so merged timelines renumber


rogvid:
Found in review of #6.

A beat's `index` is its position in its own take's `beats` list. For a
single-segment demo that is also its position in the demo. For a segmented one
it is not: segment two starts again at 0.

This becomes a real problem where two queued slices meet. #9 stores per-beat
evidence as `evidence/beat-NN.json`, named from `index`. #7 merges segment
timelines into one file. Whichever runs second has to renumber the other's
output, or the merged timeline points at the wrong evidence — and a reviewer
handed `evidence/beat-07.json` has no way to know which segment's beat 7 it
is.

## Options

- Have #7 renumber on merge and rewrite the evidence filenames with it. Keeps
  `index` meaning 'position in this file', costs a rename step.
- Give each beat a stable identity that survives merging (`segment` +
  positional index, or an opaque id), and have #9 name evidence from that.
  Costs a schema field; nothing has to be renamed, ever.

Worth settling **before** #7 and #9 both land, not after.

## Acceptance

A two-segment demo's merged timeline and its evidence files agree on which
beat is which, with no ambiguity about renumbering.

rogvid:
Settled from #9's side, in favour of **option 2** — a beat identity that
survives merging — and in a way that costs #7 nothing.

Two changes, both in the branch for #9 (`issue-9-beat-evidence`):

1. **Evidence filenames carry the segment.** `evidence_name(index, segment)`
   produces `beat-04.json` for a whole demo and `part1.seg.beat-04.json` for
   a segment, mirroring exactly how `timeline_paths()` names
   `<segment>.seg.timeline.json`. Two segments of one demo therefore cannot
   collide, whatever their indices are, and a reviewer handed
   `evidence/part2.seg.beat-07.json` can see which segment's beat 7 it is
   from the filename alone. Every file also carries `segment` and
   `beat.index` internally.

2. **The path is written onto the beat, not derived from it.** Each beat in
   `timeline.json` gains an `evidence` key holding the path relative to the
   timeline file — the same contract `still` already has:

   ```json
   { "index": 4, "verb": "shot", "still": "images/01-dashboard.png",
     "evidence": "evidence/beat-04.json" }
   ```

   Nothing anywhere rebuilds that string from `index`. `SKILL.md` says
   "read the pointer, never rebuild it", and `tests/smoke` asserts the
   pointer resolves to a file that exists and that `evidence/` holds nothing
[...]

PR #53, opened 2026-07-25
stitch() merges per-segment timelines with corrected offsets

### LINKED ISSUE #7: stitch() merges per-segment timelines with corrected offsets
Split from #1 (item 1, final bullet). Depends on the beat event log slice.

## Problem

Segmented demos record each part separately and concatenate with `stitch()`.
Each segment will produce its own `timeline.json` with timestamps relative to
its own start, so a multi-segment demo has no single coherent timeline — and
every consumer of the manifest (beat-aligned frames, AC coverage) breaks on
exactly the demos that are hardest to review by hand.

## Tasks

- [ ] `stitch()` reads each segment's `timeline.json`
- [ ] Offset each segment's beat timestamps by the cumulative duration of the
      preceding segments (`media_duration` on each `.seg.mp4`, not the
      nominal storyboard timing — encoder output is the truth)
- [ ] Write the merged `timeline.json` / `timeline.md` next to `demo.mp4`
- [ ] Account for `interlude()` beats that bracket segment boundaries

## Acceptance

A two-segment demo produces one `timeline.json` whose beat timestamps match
the stitched `demo.mp4` within ~100ms, including beats in the second segment.

### LINKED ISSUE #21: stitch() leaves orphaned .seg.timeline.json files naming deleted .seg.mp4s
Found in review of #6.

`stitch(out_dir, segments)` unlinks each `<segment>.seg.mp4` unless
`keep_parts=True`, but knows nothing about the `<segment>.seg.timeline.json`
and `.md` that #6 now writes beside them. After a stitch the directory holds
timelines whose `media` field names a file that no longer exists.

Harmless today because nothing reads them. Not harmless once #7 lands and
`stitch()` is the thing that merges them: a stale segment timeline from an
earlier run, left behind by a previous stitch, is indistinguishable from a
fresh one and would be merged into the result.

## Fix

Whatever #7 decides the merge is, the cleanup has to match `keep_parts`:
delete segment timelines exactly when their media is deleted. Best done as
part of #7 rather than before it.

## Acceptance

After `stitch(...)` with the default `keep_parts=False`, no
`*.seg.timeline.*` remains; with `keep_parts=True`, each one is still next
to its `.seg.mp4`.

rogvid:
Fixes #7. Fixes #21.

Each segment produced its own `timeline.json` with timestamps relative to its
own start, so a multi-segment demo had no coherent timeline — and every
consumer of the manifest broke on exactly the demos hardest to review by hand.

`stitch()` now loads and validates every segment's timeline **before** ffmpeg
runs (a post-concat failure would leave a fresh `demo.mp4` with no timeline),
then concatenates, merges, and cleans up. Offsets come from
`media_duration()` on each `.seg.mp4` — the encoder's answer, never nominal
beat timing.

#21: `.seg.timeline.json`/`.md` are now deleted exactly when their `.seg.mp4`
is, and kept under `keep_parts=True` since a re-stitch needs them.

## Segments were completely untested before this

`tests/README.md` listed them under Known gaps: no take recorded a segment and
`interlude` beats were unexercised. Coverage came first, merge logic second.

## The acceptance criterion, measured differentially

**+0 ms on all six takes.** The closing caption lives in segment two and is
timed twice — in `part2.seg.mp4` against that segment's own log, and in the
stitched `demo.mp4` against the merged log. `stitch` copies streams, so those
are the same frames carrying the same capture loss; the *difference* is merge
error alone, with #18 cancelled.

That cancellation forced an honest change: the harness's sharp 250ms
probes-drifting-apart bound assumes both probes rode one capture. Across a
segment boundary they do not — each segment has its own screencast and its own
~0.7s untickerable setup window, and one of four early takes lost a whole one
(−520ms). The cross-boundary bound is now one capture-loss window (750ms),
documented, with the tight claim made by the differential measurement instead
of by a bar.

## #22

Fixed by **stable identity**, not renumbering. `index` keeps meaning "position
in this file" and is renumbered on merge (timeline.md's table, "beat N (verb)"
messages and `issues[].beat` all read it that way). Every beat also carries
`segment_index`, written for every take and untouched by the merge, so
`(segment, segment_index)` names a beat identically before and after a stitch.

Renumbering was rejected because #9 writes evidence *during* a take, when
nothing knows how many segments there will be — it would make #9 depend on #7
having run, and make a re-stitch rewrite files for segments that did not
change. #22 stays open until #9 names files from that pair.

17 fault injections, each watched fail — including `nominal_timing`, the exact
bug the issue warns about (parts off by ~0.9s each, caption 2.08s from its
frame), and two anti-vacuity checks.

Filed #51: `stitch()`'s issue-merging path is untested, because the segmented
take is healthy by design.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,8 @@ __pycache__/
 .tts/
 *.seg.mp4
 # Per-segment beat logs. The merged timeline.json/timeline.md next to
-# demo.mp4 is committed; these are the parts it is built from.
+# demo.mp4 is committed; these are the parts it is built from, and stitch()
+# deletes them with their .seg.mp4 unless it was passed keep_parts=True.
 *.seg.timeline.json
 *.seg.timeline.md
 # Left behind only when a take crashes before the recorder cleans up.
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ tests/out/
 
 # Best practice
 .env
+
+# ffmpeg concat listing, written and removed by stitch()
+.concat.txt
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -550,6 +550,16 @@ at the top level carry the value every segment agrees on — `"mixed"`, and
 `null` per key, where they do not; the per-segment truth is in `segments`. A
 timeline a single take wrote has no `segments` key at all.
 
+`stitch()` refuses before it encodes anything if the parts cannot honestly be
+joined: a missing or unreadable `.seg.mp4`, a beat log of the wrong schema or
+one written for a *different recording* of that segment, or parts that
+disagree on codec, resolution, frame rate or having an audio track.
+`concat -c copy` accepts all of those and exits 0, and the damage is invisible
+afterwards — a frame-rate mismatch moves every later beat away from its frame,
+a resolution mismatch keeps the first part's dimensions, and one silent part
+makes concat drop every later part's narration. Recording every segment with
+the same `Recorder` settings is what keeps you clear of it.
+
 ## Failing the take on a broken app
 
 **A demo that looks perfect while the app throws `TypeError` on every render
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -35,7 +35,7 @@ Each demo gets one folder (suggested: `docs/guides/<YYYY-MM-DD>-<slug>/`):
 |---|---|---|
 | `record.py` | The storyboard that produced the media (re-runnable) | yes |
 | `images/*.png` | Stills captured at key moments | yes |
-| `timeline.json` | The beat log — every verb, when it ran, what caption was up | yes |
+| `timeline.json` | The beat log — every verb, when it ran, what caption was up. For a segmented demo, the merged one `stitch()` writes | yes |
 | `timeline.md` | The same log rendered for humans, stills embedded | yes |
 | `guide.md` | Optional written guide embedding the stills | yes |
 | `demo.mp4` | The recording (mp4 only — gifs get too big) | **no** — regenerate it, or attach it to the PR |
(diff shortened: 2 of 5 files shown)


rogvid:
## Cold review: BLOCK

**The 250ms → 750ms relaxation hides real merge error.** Shifting every beat
of segment *one* by 350ms: `smoke: PASSED` at 750ms, `FAILED` at 250ms. A
550ms shift also passes, reported as `capture loss, see #18`.

`check_merge_offset` times one beat — the closing caption, in the **last**
segment — so it validates that segment's cumulative offset and nothing else.
With 3+ segments the intermediate offsets have no pixel measurement at all.
"Nothing is given up" holds for the last segment only. Fix: time the
differential once per segment (pure pixel reading on files already on disk),
and restore the 250ms bar.

**Two injections miss**, on fields SKILL.md advertises as the per-segment
source of truth: `segments[].beats` hardcoded to 0, and
`segments[].recorder` hardcoded to `"Bogus"` — both pass.

**Also:** the "no timeline beside a fresh mp4" state the docstring calls
unacceptable is reachable without monkeypatching (a corrupt part makes
concat exit 0, then `media_duration` raises); the cleanup loop aborts
mid-way and re-creates #21; `stitch()` validates nothing about the streams
it joins (fps mismatch → 1.92s error, ffmpeg exit 0; a narrated part 2 after
a silent part 1 has its audio silently dropped).

### Verified sound, and under-argued in the PR

`format=duration` is the provably correct offset source for `concat -c copy`
— checked against an audio-overhang case to within one frame. Rounding does
not compound: 40 synthetic segments land at 20.28 vs a true 20.28.
`segment_index` is genuinely asserted on every take.

**#7 and #9's #22 solutions compose** — verified by merging both branches and
running `--segments-only` green, with `evidence/part2.seg.beat-00.json`
naming correctly. One gap neither noticed: `evidence_docs()` hardcodes the
merged position with no segment awareness and `run_segments` never calls it,
so the segmented × evidence interaction is still tested by nobody.

Issue #55, opened 2026-07-25
tests/README.md has duplicated paragraphs from a bad merge


rogvid:
Noticed while adding the terminal-redaction section for #5. `tests/README.md` on `main` (d667938) carries several paragraphs twice, in both versions, from merges of the parallel PRs:

- the `## Running it` code block lists `--web-only` / `--terminal-only` twice, once as "all three takes" and once as "every take" (lines ~27-37);
- `Re-running into the same --out-dir is safe:` appears twice, once naming ten take directories and once naming three (lines ~78-94);
- `Deleting is bounded.` appears twice, same split (lines ~96-103);
- `## What it asserts` opens with both "Six independent axes ... other five" and "Five independent axes ... other four" (lines ~127-130);
- the fixture hook table under `## The fixture app` lists `?console-error=1` and `?secret=1` twice, with different descriptions, and `Nothing checks audio.` appears twice in Known gaps with different content.

Each pair is "the version before a slice landed" next to "the version after". The later text is right in every case I looked at; the earlier one is stale rather than wrong-in-a-new-way.

Left alone deliberately in the #5 branch — untangling it touches the same lines three in-flight PRs are editing, and doing it during a feature slice makes that diff unreviewable. Best done as its own commit once the #7/#8/#9 queue drains.

rogvid:
Duplicate of #47, which is the same bug found in the review of #9. Line
references folded into that issue; tracking there.

Issue #54, opened 2026-07-25
addon-serialize.js is loaded into the terminal page and never used


rogvid:
While wiring the PTY scrubber for #5 I checked every path that could read screen text back out of xterm.js. There are two candidates, and one of them is dead:

- `window.__termText()` reads `term.buffer.active` directly. This is what `_screen()`, `wait_for_text()` and `wait_for_prompt()` use, and it is covered — it sees only what the scrubber wrote.
- `assets/xterm/addon-serialize.js` is injected by `TerminalRecorder._start()` (`page.add_script_tag`) and **never instantiated**. Nothing calls `new SerializeAddon.SerializeAddon()`, nothing loads it into the terminal, and `grep -rn serialize` over the repo finds exactly that one line.

So it is ~30 kB of vendored JS parsed into every terminal take for nothing. Either:

- delete the injection and the asset (and the VERSIONS.txt entry), or
- use it — a serialized dump of the terminal (ANSI intact) alongside `timeline.json` would be a genuinely useful artifact for a terminal demo, and is presumably why it was vendored in the first place.

Not urgent and not a leak (it holds no state of its own; a serializer attached later would read the same scrubbed buffer), but a reader of `_start()` reasonably assumes it is doing something.

PR #58, opened 2026-07-25
Redact secrets on the terminal PTY output path

### LINKED ISSUE #5: Redact secrets on the terminal PTY output path

rogvid:
Fixes #5.

`TerminalRecorder` pipes real shell output through a PTY into an in-browser
xterm. Anything the command prints was captured verbatim — a key echoed by a
tool, a token in a config dump, a credential in an error message. The web half
(#4) has no reach here: there is no DOM to blur.

## The "one scrub() call" assumption was wrong, and this is why

`os.read(self._fd, 65536)` returns arbitrary PTY fragments, so a secret split
across two reads is a substring of neither.

`_StreamRedactor` sits in `_pump` after the UTF-8 decoder. Per fragment it
strips SGR into a copy with an offset map, finds registered secrets and shape
matches in that copy, computes a **cut point** — the shortest suffix that could
still complete a match — masks everything before it, and carries the rest.

The cut is deliberately **not** `max(len) - 1`. A fixed window leaves a finished
screen permanently short of what the program wrote, and `wait_for_prompt` /
`wait_for_text` read that screen. It is the longest suffix that is a proper
prefix of a registered secret, or matches a shape's tail, or sits inside a match
touching the end — capped at 4096.

## Proved deliberately, not hoped for

`capped_pty_reads()` swaps the `os` module **in `demo_recording.terminal` only**
for one whose `read()` returns <= 7 bytes and keeps every fragment. The take
then asserts from those fragments that the printed key was reassembled from >= 2
— measured **5**, and the `send(Secret(...))` echo **28**. Removing the cap
makes the key arrive in **1** read on this box and that assertion fires, which
is why it exists.

## Escape interleaving: a fixed list, and the list is the reach

Matching runs against a copy with the *inert* sequences removed and matches
mapped back to raw offsets, with the escapes inside a masked run re-emitted
after the mask so colour survives. The list: SGR, mode set/reset (`\x1b[?25l`,
what every spinner emits mid-line), erase-to-EOL, BEL-terminated OSC, charset
and keypad selection. Rows `ans`, `hid` and `osc` grade three of them on the
pixels.

**It is a list, not the predicate "does not move the cursor"**, and the
difference is a leak. Measured: a CSI with an intermediate byte (`\x1b[1 q`), a
DCS string, and an ESC-aborted OSC all render the value as one word on screen
while the scrubber writes it in the clear. A *registered* value split that way
still dies at the final screen check; a shape-only token goes into the frames.
Filed as #71 and corrected in SKILL.md and the module comment.

Cursor movement is the other half and is genuinely uncovered: masking across a
jump would corrupt the redraw. **The value comes out readable, not scrambled**
— which is why the final screen check exists.

## The final screen check runs on every exit path

`_verify_redaction_final()` reads the finished terminal buffer — visible screen
*and* scrollback — and raises `SecretLeak`: no mp4, no timeline, no stills.
Three things about *when* it runs, each of which was wrong at some point in
this PR:

| | why |
|---|---|
| after the narration tail | the tail pumps a child that is still writing, so verifying first vouched for frames that are not the frames recorded |
| after `_stop()` | the scrubber withholds a trailing fragment and flushes there, so verifying first read a screen the recording does not end on |
| on **every** way out of the `with` | gated on `exc_type is None`, a wait_for_text() timeout or a Ctrl-C skipped it and kept the stills — and a terminal still is the raw screen |

Over a storyboard that already raised, the leak is printed and the original
exception propagates: a timeout replaced by a leak report costs the author the
message that says what to fix, and the artifacts are gone either way.

`terminal-cursor-leak` is three takes — `clean`, `crash`, `tail` — one per row
of that table.

## Verification

Every assertion was fault-injected and seen to fail, and the pixel assertions
grade the **rendered frames** (mp4 and PNG crops against a `[redacted]`
reference measured in the same frame), not a Python string.

Round one, 9 injections: releasing shape fragments on the first idle poll
(rows differ from the mask by 43-46); inert escapes narrowed to SGR; the
screen check stubbed out; no token-boundary rule; the fragment ceiling applied
to a registered value; the span pull-back removed; the registered-value hold
removed; AKIA and JWT shapes removed; the `key()` guard removed.

Round two, 5 more: the screen check gated back to the clean path (`crash`
keeps a 129 kB still with the key on it); the check moved back before the tail
and flush (`tail` writes an 82 kB mp4 and reports nothing wrong);
`_ANCHORED_HOLD_S` at 5.0 s (the `slw` row comes out masked and the take says
the documented limit moved); at 1.0 s (`pau` and `anc` render at 41-45 against
the mask); and the token-boundary rule removed again against the re-based lag
measurement (3.0 s against a 1.5 s bar, healthy 0.2 s).

Two assertions that **could not fail** were found and fixed rather than
trusted: `_ANCHORED_HOLD_S` was graded from one side only, so it could be set
to anything above ~0.45 s; and the held-fragment lag was timed from before the
printer ran, so it measured the script's runtime rather than the release.

## What it does not cover

Half a secret still renders if a program is killed mid-print. A registered
secret's prefix at a quiet stream is held with no timeout, so a program
printing exactly the first characters of one and waiting will appear to stall
— the recorder says so on stderr after two seconds. Only four shapes, one
spelling of each. Shape fragments are held for three seconds and no longer,
which the `slw` row now grades as a limit. Nothing sanitizes the child's
environment or its log files. And no assertion reads text back out of a frame
— the pixels prove "this row is the mask", not which mask.

Filed and not fixed here: #54, #55, #66 (line-erase and alt-screen
over-masking), #67 (teardown masking and the withholding warning are
ungraded), #71 (inert-escape under-masking), #72 (`_MAX_HOLD` releases the
head of an anchored fragment that never completes a match), #73
(`_term_deltas` compares with `zip(strict=False)`).

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -57,6 +57,23 @@ The words, and what each fixture element they belong to is for:
   TKCT  #token-control     a field's control — same chrome, never redacted
   SDCT  #sd-token-control  the same, inside the shadow root
   LATE  a caption          set before the value in it was registered
+
+The terminal half of redaction (issue #5) needs four more, and they are not
+all `sk-live-` shaped because the shapes themselves are what is being tested:
+
+  PTYK  sk-live-…   registered, printed by a command under a PTY
+  ANSI  sk-live-…   the same, printed with a colour escape inside it
+  SHAP  ghp_…       registered by nobody — the shape net is all that hides it
+  SEND  pw-live-…   sent as Secret(...); deliberately matches no shape, so
+                    that what masks it can only be the registration
+  TCTL  zz-live-…   never registered and matching no shape, on purpose: it is
+                    the legible reference every masked row is measured against
+
+Same convention as above — a four-letter word and a run of zeroes — and the
+same reasoning: the default ruleset flags none of them (the GitHub rule wants
+exactly 36 characters after `ghp_`; this one has 24), and these entries exist
+so a future release that does start flagging them cannot turn the job red for
+strings that were never credentials.
 """
 paths = ['''^skills/demo-video/helpers/assets/''']
 regexes = [
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -58,22 +58,40 @@ The words, and what each fixture element they belong to is for:
   SDCT  #sd-token-control  the same, inside the shadow root
   LATE  a caption          set before the value in it was registered
 
-The terminal half of redaction (issue #5) needs four more, and they are not
-all `sk-live-` shaped because the shapes themselves are what is being tested:
+The terminal half of redaction (issue #5) adds a family of its own, and they
+are not all `sk-live-` shaped because the shapes themselves are what is being
+tested. **No allowlist entry was added for any of them, and none is needed** —
+the `regexes` list below covers the web literals only. They are described here
+because this file's comments are the audit trail for every fake credential in
+the repo, and a value nobody wrote down is a value the next person has to
+re-derive:
 
   PTYK  sk-live-…   registered, printed by a command under a PTY
   ANSI  sk-live-…   the same, printed with a colour escape inside it
+  HIDE  sk-live-…   the same, cut by a cursor-hide escape
+  OSCT  sk-live-…   the same, cut by a window-title escape
+  PAUS  sk-live-…   the same shape, paused mid-value inside the hold window
   SHAP  ghp_…       registered by nobody — the shape net is all that hides it
+  STRM  ghp_…       the same, written one character at a time
+  ANCH  ghp_…       the same, paused after its first character
+  SLOW  ghp_…       the same, paused *past* the hold window — the one value in
+                    the take that is supposed to render in the clear, because
+                    it grades a documented limit
+  CURS  sk-live-…   written at two cursor positions, which must kill the take
+  KEYD  sk-live-…   spelled out through key(), which must be refused
+  AKIA…             an AWS-shaped id nobody registered
+  eyJ….eyJ….        a JWT-shaped token nobody registered
   SEND  pw-live-…   sent as Secret(...); deliberately matches no shape, so
                     that what masks it can only be the registration
   TCTL  zz-live-…   never registered and matching no shape, on purpose: it is
                     the legible reference every masked row is measured against
 
 Same convention as above — a four-letter word and a run of zeroes — and the
-same reasoning: the default ruleset flags none of them (the GitHub rule wants
-exactly 36 characters after `ghp_`; this one has 24), and these entries exist
-so a future release that does start flagging them cannot turn the job red for
-strings that were never credentials.
+same reasoning: the default ruleset flags none of them. The GitHub rule wants
+exactly 36 characters after `ghp_` and these have 24; the AWS rule wants
+exactly 16 after `AKIA` and this has 24; the JWT rule wants 17 or more in each
+of the first two segments and needs the second to begin `ey`, and this one has
+neither. Verified by running gitleaks over the whole tree, not assumed.
 """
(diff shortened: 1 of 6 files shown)


rogvid:

Four secrets fully legible in stills and mp4 frames, **two of them
registered**, with `issues: []` and `strict=True` passing.

### Blocking

1. **Shape tokens leak at a quiet boundary** — `_hold_from` applies the
   shape rules only when `mode == "read"`, and `quiet()` writes incomplete
   fragments out unmasked. **No `sleep` required**: measured by write rate,
   0 and 1 ms/char mask, **5 ms/char leaks the whole token**, as do 20 and
   100 ms. That is ordinary streaming output and every typewriter demo. This
   is #5's own acceptance criterion.
2. **Registered secrets split by non-SGR escapes that do not move the
   cursor** — `\x1b[?25l` (every spinner library), `\x1b[K`, OSC title
   sequences. Non-contiguous in the stream, perfectly contiguous on screen,
   fully legible.
3. **The documented rationale for the uncovered case is false.** A
   cursor-addressed repaint renders *contiguous and legible*; SKILL.md tells
   the reader it "is not contiguous on screen". Conversely, a
   terminal-wrapped line is listed as uncovered and **is** caught. Wrong in
   both directions.
4. **No terminal `_verify_redaction_final`.** The web half has a last-word
   check; this has none, which is why every leak above wrote a normal mp4. A
   one-line `secret in self._screen()` at teardown catches 2 and 3.

### The test cannot reach the bug

`capped_pty_reads` forces *read* splits — the easy half — and structurally
cannot force a *quiet* split, since the printer emits every line back-to-back
so `select()` never idles mid-token. `TERM_HELD_LINE`, the only thing
exercising `quiet()`, asserts the release that **causes** finding 1.

### Also

Three injections pass, not one — `_mask_dangling`, the `sk-` shape, and the
`start < cut < end` clamp fixup, the last admitted nowhere. `key()` writes
the secret into `timeline.json`/`timeline.md`, the files SKILL.md says to
commit. The "held forever" hazard is a *silently truncated screen*, and
`_mask_dangling` will replace an innocent 4-char word (`deploy-`, `prod-`)
with `[redacted]` in the final frames.

### Verified sound

The mask genuinely reaches the pixels, with properly calibrated bars.
Mechanism isolation is real. The read cap is load-bearing. A 5008-char
registered secret masked whole. `ls` output, `\r` repaints, right-margin
wrap and `send(Secret)` all held.

PR #113, opened 2026-07-26
The transition in and the transition out

### LINKED ISSUE #91: A card interlude over TerminalRecorder is never cleared, and every artifact except the frames says the demo worked
Found while running the skill on its own use case (#64), on the first take of
the reference demo. It cost the whole terminal half of a 60-second video, and
**nothing except the review frames noticed.**

## What happened

The storyboard opened its terminal segment the way SKILL.md's Process step 1
describes opening a new segment:

```python
with TerminalRecorder(HERE, segment="part2") as rec:
    rec.interlude("The same filter, on the command line.")
    rec.caption("The CLI reads the same queue.")
    rec.run("./tickets list")
    ...
```

`interlude()` with the default `style="card"` paints a full-screen title card.
Nothing clears it — `interlude("")` is the only thing that does, and the
storyboard never called it. On the **web** recorder the next `goto()` destroys
the card with the document, which is why this has never bitten; a
`TerminalRecorder` never navigates, so the card stayed up for the entire
segment.

Result: **24.3 s of a 60.2 s video — 40% — is a static title card**, with the
terminal running underneath it, invisible. Verified directly:
`ffmpeg -ss 50 -i demo.mp4 -frames:v 1` is the card.

## Why this is worse than an authoring mistake

Every artifact the skill produces for a reviewer, except one, reported success:

| Artifact | What it said |
|---|---|
| stderr summary | `1 problem(s) recorded` — the deliberate non-zero exit. Nothing about the card. |
| `timeline.json` / `.md` | all 17 terminal beats, in order, with `exit_code: 0` on the two commands that succeeded |
| `evidence/part2.seg.beat-*.json` | `screen` holds the **full CLI output**, ticket rows and all — the terminal buffer was correct; it was just covered |
| `images/05-cli-list.png` … `07-cli-unknown.png` | the card |
| `frames/beat-26..42.png` | the card — 17 consecutive identical frames |

A reviewing agent given the acceptance criteria plus `evidence/` and
`timeline.md` (no video) **approved the CLI criterion as demonstrated**, quoting
`evidence/part2.seg.beat-10.json` → `screen`. It was reading the truth about the
program and inferring something false about the recording. The fresh-eyes frame
reviewer caught it immediately: *"Seventeen consecutive frames — over 20 seconds,
a third of the video — are pixel-identical. No terminal, no prompt, no command."*

That is the failure shape #62 calls out: an artifact that is confidently wrong is
worse than one that is absent. `screen` is not a claim about what was **visible**,
and nothing in the file says so.

## Suggestions, roughly in order of how much they cost

1. **Make the take say so.** A segment whose frames never change is detectable
   without any new machinery — scene-change detection already runs inside beats
   ≥3 s. A take that recorded N seconds with no scene change at all should print
   it in the end-of-take summary, next to the issue list.
2. **Clear a card when a verb acts on what it covers.** `run()`, `send()`,
   `key()` and `shot()` on a `TerminalRecorder` all address a screen the card is
   hiding. Either clear it implicitly or refuse.
3. **Say it in SKILL.md.** The Process section shows `interlude()` opening a
   segment and never shows it being cleared, because on the web path the
   following `goto()` does it. The terminal path has no `goto`.
4. **Have `evidence` state what it is not.** `screen` and `aria` describe the
   buffer and the DOM, not the picture. One line in the Per-beat evidence
   section, and ideally a field, would have stopped a reviewer certifying a
   video it could not see.

Suggestion 1 is the one that generalises: any full-screen overlay left up — a
modal, a light interlude, a spotlight scrim — produces exactly this, and the
recorder is the only thing positioned to notice.

Measured on: `examples/ticket-queue` (#89), take of 2026-07-26, 43 beats,
60.2 s, 2 segments (`Recorder` + `TerminalRecorder`), stitched.

### LINKED ISSUE #110: A terminal segment opening with interlude() shows ~290ms of bare terminal first
Found by a human watching the reference demo (#64) — the first defect in this
project caught by the video review step rather than by a test or a code reviewer.

## What it looks like

At the web→terminal segment boundary the viewer sees: web app → **a flash of
bare terminal** → the transition card → terminal again. It reads as a glitch.

## Measured

From the reference demo's merged `timeline.json`:

```
part1 (Recorder)          duration 37.76   offset  0.00
part2 (TerminalRecorder)  duration 23.40   offset 37.76
   38.05  interlude   "The same filter, on the command line."
```

The terminal segment's first frame lands at **37.76s**; the interlude card does
not paint until **38.05s**. That is **~290ms of bare terminal** — an empty
window with a lone prompt, confirmed by extracting the frame at 37.9s.

## Cause

`TerminalRecorder.__enter__` starts capturing as soon as the terminal page is
up. A storyboard cannot paint the card any earlier than its first statement
inside the `with`, so the gap is structural: it is the cost of starting the
recorder before the storyboard's first verb, and it exists for every segment
that opens with an `interlude()`.

The web recorder does not show it because `Recorder.__enter__` has nothing on
screen until `goto()` — there is no "before" to flash.

## Directions

- **Hold the first frame until the first beat.** Have `TerminalRecorder`
  suppress the composited output (or paint the interlude backdrop) until the
  storyboard's first verb runs, so the segment opens on intent rather than on an
  empty prompt.
- **Or open segments on a card by construction** — an `interlude=` constructor
  argument painted before capture starts, which also removes the ordering
  footgun that produced #91.

The second is likely better: it makes "a segment opens with a title card" a
property of the recorder rather than something every storyboard must remember to
do first, and #91 showed the manual version has a second sharp edge.

## Acceptance

A terminal segment that opens with an interlude shows the card in its first
frame. No bare-terminal frames before it.

## Related

#91 — the same card, never *cleared*, on the previous take of this same demo.
Different bug, same seam: `interlude()` and `TerminalRecorder` do not fit
together at either end.

### LINKED ISSUE #111: spotlight() eases in but snaps out: the clear wipes the transition property in the same frame as the transform
Found by a human watching the reference demo (#64). The spotlight eases *in* over
250ms and **snaps out instantly**. The asymmetry is what reads as wrong — a
smooth grow followed by an abrupt pop back to normal.

## Cause — it is one line, and it is exact

`web.py:160-176`:

```js
window.__demoSpotlight = (el) => {
  window.__demoSpotlightClear();
  window.__spotEl = el;
  window.__spotPrev = el.getAttribute('style') || '';
  el.style.transition = 'all .25s ease';      // <- transition lives in the style attr
  el.style.outline = '3px solid rgba(ACCENT,.85)';
  el.style.transform = 'scale(1.02)';
  ...
};
window.__demoSpotlightClear = () => {
  if (window.__spotEl) {
    window.__spotEl.setAttribute('style', window.__spotPrev);   // <- wipes it
    window.__spotEl = null;
  }
};
```

The clear restores the element's **whole original style attribute** in one
assignment. That removes `transition: all .25s ease` in the *same frame* as it
removes `transform` and `outline` — so there is nothing left to animate the exit
with. The enter transition is real; the exit has no transition property at the
moment it would need one.

Restoring `__spotPrev` verbatim is otherwise correct and worth keeping: it is
what guarantees the element is returned exactly as it was found, including an
inline style the app itself set.

## Direction

Revert the spotlight properties individually while the transition is still on the
element, then restore `__spotPrev` on `transitionend` (with a timeout fallback,
because `transitionend` does not fire if nothing actually changed):

```js
window.__demoSpotlightClear = () => {
  const el = window.__spotEl;
  if (!el) return;
  const prev = window.__spotPrev;
  window.__spotEl = null;
  el.style.outline = ''; el.style.outlineOffset = '';
  el.style.background = ''; el.style.transform = '';
  const done = () => el.setAttribute('style', prev);
  el.addEventListener('transitionend', done, {once: true});
  setTimeout(done, 400);
};
```

Two constraints on whoever picks this up:

- **`spotlight()` must stay synchronous from the storyboard's point of view.**
  The clear currently completes before the next verb; an exit animation means the
  next beat can start mid-fade. Decide whether the verb waits out the transition
  (simplest, costs 250ms per clear) or whether overlapping is acceptable.
- **`deterministic=True` flattens CSS transitions**, so under determinism both
  directions will snap. That is correct and consistent — but it means the fix is
  invisible in any deterministic take, including most of `tests/smoke`. Grade it
  in a non-deterministic take or the assertion cannot fail.

## Acceptance

The spotlight leaves an element over the same duration and easing it arrived
with, and the element's inline style is byte-identical to what it was before the
spotlight once the transition has finished.

rogvid:
Two defects a human found by watching the reference demo — the first in this
project caught by the video review step rather than by a test. Both are
demoable by this repo's own rule, so **this PR has a demo and the video is what
should be reviewed**; the numbers below are here so the reviewer can check the
video against them, not instead of it.

## #111 — the spotlight eased in and snapped out

`__demoSpotlightClear` restored the element's whole `style` attribute in one
assignment. That is correct and worth keeping — it is what guarantees the
element is handed back exactly as it was found, including an inline style the
app itself set — but it removed `transition` in the *same frame* as it removed
`transform` and `outline`, so there was nothing left to animate the exit with.

The spotlight's own properties are now reverted individually while the
transition is still on the element, and the wholesale restore waits for
`transitionend`. Three things about that:

- **The ring fades by animating its colour to alpha 0**, not by clearing the
  `outline` shorthand. `outline-style` is not animatable, so clearing it makes
  the ring vanish in one frame while the scale eases — half a snap.
  `background` and `transform` *are* cleared outright, because removing an
  inline declaration transitions to the value underneath it, which is the value
  the element is supposed to return to.
- **The `transitionend` fallback timer is sized from the element's own
  computed transition time**, not from a constant. `deterministic=True`
  flattens transitions to 1 ms with `!important`, so under determinism the
  exit is over in a millisecond and the timer never runs. A hardcoded 400 ms
  would have charged every deterministic take — `tests/smoke` included —
  nearly half a second per clear for a transition that had already finished.
- **An element that had no `style` attribute gets none back.** `__spotPrev`
  was `getAttribute('style') || ''`, so a clear left `style=""` behind. That is
  invisible in every frame and is exactly what the acceptance criterion's
  "byte-identical" asks about.

### The decision the issue asked for: the verb waits

`spotlight()` blocks until the exit transition has finished and the style
attribute is back. Two reasons and one measurement:

- **Two consumers read that attribute.** This beat's evidence records the
  spotlight target's `outerHTML` *including* `style`, and `redact()`'s
  MutationObserver re-asserts the mask off the same attribute. A verb that
  returns mid-fade makes both of them read a value that depends on when the
  compositor happened to fire. A storyboard cannot be written against that,
  and an artifact that varies run to run is the kind this repo refuses.
- **Overlapping buys nothing here.** The clear runs *before* the new entrance,
  so "overlap" would mean two elements half-lit at once, which is worse to
  watch, not better.

What it costs, measured end to end including the verb's existing `pause(0.3)`:

| | before | after |
|---|---|---|
| `deterministic=False` | 0.31 s | **0.55 s** |
| `deterministic=True` | 0.31 s | 0.33 s |

## #110 — a terminal segment opening on ~290 ms of bare terminal

`TerminalRecorder(interlude="…")`, painted before capture starts, rather than
suppressing frames or leaving it to the storyboard.

I went with the constructor argument over "hold the first frame until the first
beat" for the reason the issue gives — it makes *a segment opens on a card* a
property of the recorder rather than something every storyboard must remember —
plus one the issue implies: the recorder that raises the card is then the one
that can clear it, which closes #91's end of the same seam. There is no
ordering left for a storyboard to get wrong.

Mechanism: a **context init script**, which runs before the page's own scripts
on every document including Chromium's initial one. That is earlier than any
Python statement can reach — earlier than `pty.openpty()`, earlier than
xterm.js is injected — and it survives `_start()`'s own `goto("about:blank")`,
which an `evaluate()` would not. The card is appended **already opaque**;
`__demoInterlude` creates it at `opacity: 0` and raises it, which is a 450 ms
fade, and a card that fades in over the recorder's setup is showing exactly
what it exists to cover.

`interlude()` is unchanged, #91's clearing behaviour is unchanged, and the
opening card records an ordinary `interlude` beat so `timeline.json` reads the
same whichever raised it. `Recorder` (web) deliberately does **not** get the
argument — see stated limits.

## How it is graded

Two takes in `tests/smoke`, reachable as `--polish-only` and from the medium
flags. Both grade pixels, from the viewer's side, importing nothing from the
recorder.

`spotlight/` records with `deterministic=False` — the only take in the file
that does, and it has to: the determinism rule flattens the transition the
assertion is about, so under it both directions snap, both are correct, and the
assertion could not fail for the reason it claims.

**A transition is measured as the states it passes through.** Every frame of
the element's crop is scored by how far it is from the settled frame at the end
of its window, as a fraction of how far the frame at the start is; a frame
between 12% and 88% is one the transition was part way through. A snap produces
**none** — not "few", because there is no intermediate state to be in.

| intermediate frames | enter | exit |
|---|---|---|
| eased, four takes | 3-4 | 3-4 |
| the pre-#111 clear, injected | 3-4 | **0** |
| recorded with `deterministic=True` | **0** | **0** |

The first version of this counted frames that *changed*, which is not the same
claim: a hard cut in H.264 rings for a frame or two afterwards, so a snap also
produces three consecutive frames that differ from each other and the bar had
to sit uncomfortably close to them. It was also unstable — an `fps=` filter
whose grid did not share the video's phase duplicates frames, and a duplicated
frame is indistinguishable from one in which nothing moved. Two readings of the
same healthy take differed by half before the sweep switched to native rate.

`terminal-opening/` reads one corner of the frame, outside the terminal window,
at 20 fps for the whole take. The card is `#1c1a17` full-bleed (26 mean luma);
bare terminal is the recorder's pastel gradient (226). Neither is a threshold
anybody tuned — they are two constants in the recorder's own styling, an order
of magnitude apart, and the corner is read off the live `#__term_win` box.
Three statements: the **first** frame is card; the card covers at least the
first second (it is held 2.5); and the corner **does** end up bare, which is
the control — without it the first two are equally true of a recorder that
painted a black rectangle and stopped, which is #91.

## Fault injection

Nine, each one exact string in one file, through a driver that refuses to run
unless the pattern matched **exactly once** and prints the diff before running.

| injected | fails with |
|---|---|
| the pre-#111 clear restored back | `clearing the spotlight passed through 0 intermediate frame(s) [], under the 2 bar, while the enter passed through 4` |
| individual reverts kept, attribute restored in the same frame | `…passed through 0 intermediate frame(s) [], …while the enter passed through 3` |
| record the take with `deterministic=True` | `the *enter* transition passed through 0 intermediate frame(s) … either the spotlight is not being drawn, or this take lost its deterministic=False` |
| the enter's `transform` removed | `spotlight('#kpi-rev') left the element's computed transform at 'none' — the highlight never went on, so the exit measurement below is about nothing` |
| the clear leaves `style=""` | `#kpi-rev's inline style is (has_attr, value) [True, ''], but it was [False, None] before the spotlight` |
| the clear's promise resolves early | `spotlight() cleared and returned in 0.32s, under the 0.45s bar — it did not wait out its own exit transition` |
| the card raised 400 ms into the document | `the recording's first frame reads 226.5 mean luma … which is 'bare' and not the card` + `the card covers only the first 0.00s` |
| the card built at `opacity: 0` | the same two, plus the content floor |
| the card never cleared | `the corner never reaches 150.0 mean luma in 12.8s — the card the recorder raised is never taken down (issue #91). Corner range 25.8..27.0` |

**One of these found a real hole in my own work.** The clear-resolves-early
injection *passed* a version of this take whose only non-pixel assertion
compared the element's inline style before and after. It passed because the
verb ends in a 300 ms pause, so a style read after the verb is 300 ms too late
to see a restore that was still pending when it returned — and the frames are
identical either way, because the transition does animate.
`MIN_SPOTLIGHT_CLEAR_S` is the assertion that exists because of it, and a lower
bound on a wait is also the one kind of timing bar contention cannot turn red.

Full `tests/smoke` green locally (all phases, including the segments take that
uses `interlude()` and the redaction take whose observer assertion is about
`spotlight()`'s style rewrite). Recorded on an idle 16-core box with the suite
lock held; `--polish-only` run four times for stability before the bars were
set.

## Stated limits

- **A card that *fades in* over 450 ms is not caught.** Measured: a faded-in
  card and a painted one produce byte-identical corner readings for the first
  14 frames, because the recorder spends longer than the fade injecting
  xterm.js before Chromium's screencast emits anything. The recorder has no
  fade-in path — the element is appended already opaque, so no transition can
  run — but the suite is relying on setup cost for that and would not notice if
  it changed. `tests/README.md` says so.
- **`Recorder` (web) gets no `interlude=`.** It does not have the defect: its
  page is blank until `goto()`, so there is no "before" to flash — and a card
  painted before that first `goto()` would be destroyed by it, so the mechanism
  would have to be a different one solving a problem that does not exist.
- **`spotlight()` is 0.24 s slower per clear** on a non-deterministic take.
  Deliberate, argued above, and ~0 under `deterministic=True`.
- **The exit's easing is not asserted to be the *same* easing as the
  entrance**, only that both pass through intermediate states over comparable
  windows. Both read the one `SPOTLIGHT_TRANSITION` constant, so they cannot
  differ without somebody editing the constant, but no assertion says so.

Fixes #110
Fixes #111

### DIFF
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -318,7 +318,7 @@ recorded a console error, an uncaught exception, or a non-zero exit — see
 | `move_to` / `click` / `click_fast` / `scroll_to` | Visible cursor motion; `click_fast` for elements that re-render continuously |
 | `type_into(selector, text)` | Click a field and type visibly, key by key — form demos (checkout, login, search) |
 | `wait_for(selector)` | Wait for something the app does on its own |
-| `spotlight(selector)` | Ring + enlarge the element the caption discusses; `spotlight()` clears |
+| `spotlight(selector)` | Ring + enlarge the element the caption discusses; `spotlight()` clears. It eases in *and out* over 250 ms, and the verb waits out its own exit, so the element is back exactly as it was found before the next beat starts (~250 ms per clear on an ordinary take, ~0 under `deterministic=True`, which flattens the transition) |
 | `terminal(cmd)` / `terminal_output(text)` / `terminal_close()` | A *decorative* on-screen terminal card for off-browser actions **inside a web demo** — a prop, not a real shell. To record an actual CLI/TUI use `TerminalRecorder` (below). |
 | `redact(*selectors)` | Blur these elements for the whole take — frames and stills. **Plain CSS selectors only.** Call it **before** the first `goto()`. See **Redacting secrets**. |
 | `register_secret(*values)` | Register literal text that must never be captioned, spoken, or logged. A caption containing it raises `SecretLeak`. **Refuses values under 8 characters** — see **Redacting secrets**. |
@@ -1256,6 +1256,34 @@ with TerminalRecorder(Path(__file__).parent) as rec:
 | `wait_for_text(pattern, timeout_s=60)` | **The universal sync.** Wait until the rendered screen (visible text + scrollback, ANSI-stripped) matches `pattern`; `^`/`$` anchor to screen lines. |
 | `rec.page` | The live Playwright page (escape hatch). `rec._write(str)` sends raw bytes to the PTY. |
 
+### Opening a terminal segment on a title card
+
+**Pass `interlude=` to the constructor — do not make `interlude()` the
+storyboard's first statement.**
+
+```python
+with TerminalRecorder(
+    out_dir, segment="part2",
+    interlude="…the same thing, on the command line.",
+    interlude_hold=2.8,          # optional; the default
+) as rec:
+    rec.caption("Same filter, one command.")
+    rec.run("orders --city Berlin")
+```
+
+The recorder starts capturing when it creates the page, which is before any
+storyboard statement can paint anything — so a card raised by the first verb
+arrives ~290 ms late and the segment opens on an empty terminal with a lone
+prompt. `interlude=` raises the card from a context init script instead,
+before the page has painted at all, and the PTY, the terminal's own setup and
+the shell's first prompt all happen behind it. The recorder also **takes it
+down** when `interlude_hold` is up, so there is no `interlude("")` to forget.
+
+It records an ordinary `interlude` beat, so `timeline.json` reads the same
+either way. `Recorder` (web) has no such argument and does not need one: its
+page is blank until `goto()`, so there is no "before" to flash — and a card
+painted before that first `goto()` would be destroyed by it.
+
 ### Driving the four patterns
 
 - **Non-interactive CLI:** `run(cmd)` → `wait_for_prompt()`.
@@ -1309,7 +1337,9 @@ with TerminalRecorder(Path(__file__).parent) as rec:
    `Recorder(out_dir, segment="part1")`, poll between segments until the
    work is done, open the next segment with `rec.interlude("…a few minutes
    later…")` on `about:blank` before navigating, and `stitch()` into
-   demo.mp4. `stitch()` also merges the segments' beat logs into one
+   demo.mp4. A **terminal** segment takes its opening card as
+   `TerminalRecorder(interlude="…")` instead — see **Opening a terminal
+   segment on a title card**. `stitch()` also merges the segments' beat logs into one
    `timeline.json` / `timeline.md`, so a segmented demo commits the same two
    files as any other — you do not have to do anything for that.
 2. **Write `record.py`** in the demo folder as a short storyboard. Capture
--- a/skills/demo-video/helpers/demo_recording/core.py
+++ b/skills/demo-video/helpers/demo_recording/core.py
@@ -67,19 +67,28 @@ window.__demoCaption = (text) => {
 
 # Interlude: a full-screen title card for jumps over real-world time the
 # recording skips (minutes of background work between two segments).
+#
+# The card's styling is a module constant because a *second* thing builds this
+# same element: a segment that opens on a card raises it from an init script,
+# before the recorder's own setup and before the page has painted at all
+# (issue #110, and see terminal.py's `_OPENING_CARD_JS`). Both have to produce
+# an element `__demoInterlude` will then recognise and fade out, which means
+# one id and one stylesheet, in one place.
+INTERLUDE_ID = "__demo_interlude"
+INTERLUDE_CSS = (
+    "position: fixed; inset: 0; display: flex; align-items: center;"
+    " justify-content: center; background: #1c1a17; color: #f7f4ee;"
+    " font: 500 30px/1.5 system-ui, sans-serif; text-align: center;"
+    " padding: 0 12%; z-index: 2147483647; opacity: 0;"
+    " transition: opacity .45s ease; pointer-events: none;"
+)
 _INTERLUDE_JS = """
 window.__demoInterlude = (text) => {
-  let el = document.getElementById('__demo_interlude');
+  let el = document.getElementById('__ID__');
   if (!el) {
     el = document.createElement('div');
-    el.id = '__demo_interlude';
-    el.style.cssText = `
-      position: fixed; inset: 0; display: flex; align-items: center;
-      justify-content: center; background: #1c1a17; color: #f7f4ee;
-      font: 500 30px/1.5 system-ui, sans-serif; text-align: center;
-      padding: 0 12%; z-index: 2147483647; opacity: 0;
-      transition: opacity .45s ease; pointer-events: none;
-    `;
+    el.id = '__ID__';
+    el.style.cssText = '__CSS__';
     document.body.appendChild(el);
   }
   el.textContent = text;
@@ -1898,7 +1907,10 @@ class _DemoBase:
                 _CAPTION_JS.replace("__CAPFONT__", str(self._caption_font_px))
                 .replace("__CAPBOTTOM__", str(self._caption_bottom_px))
             )
-            self._context.add_init_script(_INTERLUDE_JS)
+            self._context.add_init_script(
+                _INTERLUDE_JS.replace("__ID__", INTERLUDE_ID)
+                .replace("__CSS__", INTERLUDE_CSS)
+            )
             self._context.add_init_script(_BRIDGE_JS)
             # Medium-specific init scripts (cursor, spotlight, ...).
             self._init_context(self._context)
--- a/skills/demo-video/helpers/demo_recording/terminal.py
+++ b/skills/demo-video/helpers/demo_recording/terminal.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import codecs
 import fcntl
+import json
 import os
 import pty
 import re
@@ -35,7 +36,16 @@ from collections import deque
 from collections.abc import Callable, Sequence
 from pathlib import Path
 
-from .core import SECRET_MASK, Secret, SecretLeak, _beat_verb, _DemoBase, _env
+from .core import (
+    INTERLUDE_CSS,
+    INTERLUDE_ID,
+    SECRET_MASK,
+    Secret,
+    SecretLeak,
+    _beat_verb,
+    _DemoBase,
+    _env,
+)
 
 _ASSETS = Path(__file__).parent.parent / "assets" / "xterm"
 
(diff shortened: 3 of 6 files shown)


rogvid:
CI's own readings on the 2-core runner, for the "threshold tuned to this box" question — same numbers as the 16-core box the bars were set on:

```
smoke: spotlight demo.mp4 ok (7.4s, 100 kB, content 19.1)
smoke: spotlight eases both ways (enter 3 intermediate frames, exit 3)
smoke: terminal-opening demo.mp4 ok (12.9s, 114 kB, content 4.6)
smoke: terminal-opening card covers the first 2.85s (corner 26 -> 226), then clears
```

Bars are 2 intermediate frames and 1.0 s of card.

Follow-up filed as #114: `interlude()` as a terminal take's first beat still records the flash, and nothing warns.

Issue #17, opened 2026-07-25
Expose the recorder's frame geometry as public API so tests need not read _geom


rogvid:
`tests/smoke` has to know where the app sits inside a recorded frame, because
scoring the whole frame does not work — a fifth to a third of it is the
recorder's own chrome, whose contrast dominates any measurement. (A fully blank
web recording scores 61.8 on a whole-frame metric; a healthy one scores 60.2.)

It currently gets that rect by reaching into recorder internals:

```python
g = rec._geom                       # web: the composited window position
box = rec.page.locator("#__term_host").bounding_box()   # terminal
```

Reading it at record time rather than re-deriving the formula is deliberate and
already much better than hardcoding: a change to the window geometry carries the
measurement with it, and deleting `_geom` fails loudly. But it is a private
attribute and an internal element id, and a change that keeps `_geom` while
moving the app elsewhere in the frame would silently make the content check
score the wrong pixels — the failure mode being that the check stops being able
to fail, which is the one nobody notices.

## Proposal

Give both recorders a small public accessor, e.g.

```python
rec.content_rect()   # (x, y, w, h): where the recorded subject sits in a frame
```

- `Recorder` returns the composited app rect (what `_frame_geometry()` already
  computes).
- `TerminalRecorder` returns the xterm host box.

That is one method each, it is genuinely useful beyond tests (anything wanting
to crop, blur, or annotate a frame needs it — see #4 on redaction, which has the
same "where is the content" problem), and it lets `tests/smoke` drop both
private accesses.

Documented as a known gap in `tests/README.md`.

rogvid:
#4 (web redaction) landed and adds a third private read of the same kind, so
this is now three:

```python
g = rec._geom                                            # web video content rect
box = rec.page.locator("#__term_host").bounding_box()    # terminal
info = {"geom": dict(rec._geom), "size": (rec._size["width"], ...)}  # redaction take
```

The redaction take needs it for a different reason than the content checks do —
it maps a *page-space* element rect into frame coordinates (`to_video_rect()` in
`tests/smoke`) to measure whether a blurred element is legible in demo.mp4 —
but it needs exactly the numbers `_geom` holds, plus the viewport, and it
re-derives the scale factor from them.

So `content_rect()` as proposed would not quite serve it: what the redaction
check wants is a *mapping*, `page_rect -> frame_rect`, of which `content_rect()`
is the special case for the whole page. Worth considering that shape when this
[...]

PR #120, opened 2026-07-27
Measure the picture, not only the storyboard (#97)

### LINKED ISSUE #97: Nothing verifies the recording shows anything: every artifact describes the storyboard, not the frames
Generalised from #91, which is the specific instance. #91's fix (clear the card)
does not close this.

## The finding

**Every artifact demo-video produces describes what the storyboard *did*. None
describes what the frames *show*.** So a recording can be entirely occluded,
blank, or unstyled, and every artifact will agree it succeeded.

Measured on the first take of the reference demo (#64). A title card covered the
terminal for **24.3s of a 60.2s video** — 40% of the recording, with the real
content running underneath it, invisible:

| Artifact | What it reported |
|---|---|
| stderr summary | `1 problem(s) recorded` — the deliberate non-zero exit. Nothing about the card. |
| `timeline.json` / `.md` | all 17 terminal beats, in order, correct `exit_code`s |
| `evidence/*.json` | `screen` held the **full CLI output** — the buffer was correct, it was just covered |
| `images/*.png` | the card |
| captions, narration, stills | all consistent with a healthy take |

Only `frames/` disagreed — 17 consecutive identical PNGs — and only because
somebody looked at the pictures.

## Why this is the sharpest version of "the artifact must never lie"

Two consequences, both measured rather than argued:

1. **A reviewing agent given the criteria plus `evidence/` and `timeline.md`
   approved the CLI criterion as demonstrated**, correctly quoting the evidence.
   It was reading the truth about the *program* and inferring something false
   about the *recording*. Evidence describes the buffer; nothing anywhere states
   that this is not a claim about the picture.
2. **The beat table for that take reads as a complete, successful CLI demo** —
   17 beats, exit codes, captions in order. That is the "tier 1" text artifact
   from #2, and it was confidently wrong about whether anything was visible.

Under #62's bar — *never stale, never silently partial, never confidently
wrong* — this is the worst case in the project: an artifact set that
unanimously certifies a demo nobody can watch.

## What exists, and why it does not cover this

`tests/smoke` has a content floor (`gray_frames()` plus luma standard deviation
over the app rect). It is a **test-suite** check, not a **product** check — a
user recording their own demo gets nothing. Its limits are recorded in #63 via
#16: the floor sits below a total stylesheet failure, and the recorder's own
chrome dominates any whole-frame metric, which is why #17 exists at all.

## Direction

The recorder already knows where the app sits inside the composited frame.

- **Sample N frames and score the content rect**, not the whole frame — the same
  measurement the suite makes, moved into the recorder.
- **Report identical-frame runs.** 17 consecutive identical frames over 20s is
  the signal that caught this, and it is nearly free alongside `scene_times()`,
  which already runs.
- **Put it in `timeline.json`** so the text tier stops being confidently wrong:
  a `content_score` or `static_for` field a reviewer — human or agent — can see
  without opening the video.

Warn rather than refuse, at least initially. A demo legitimately holding one
frame through a long beat is ordinary, and failing a take on a heuristic is
worse than the heuristic missing — the same argument that scoped the terminal
verifier to registered values only in #5.

## Acceptance

A take whose content rect never changes for a long stretch, or never rises above
the blank floor, says so — in the stderr summary and in `timeline.json` —
without anyone opening the video. Take 1 of the reference demo would have been
caught by it.

## Deliberately not included

Judging whether the demo shows the *right* thing. That is #12's job, and a
human's. This is only: did the recording capture anything at all.

rogvid:
Fixes #97.

## Round 2: the held-picture arm was a timer, not a detector

Review found it firing on ordinary healthy demos on both media, with a message
that named a cause which had not happened. Both are fixed; the detail is in
[the round-2 comment](#issuecomment). Short version:

| take | `static_for` |
|---|---|
| healthy terminal, 2 touring captions | 16.5s |
| healthy web, 3 touring captions | 20.0s |
| healthy terminal, 3 touring captions | 21.5–22.0s |
| **reference take 1, card over the terminal** | **23.0s** |

No threshold separates those. The stretch is now correlated with the beat log:
a held picture spanning only `caption`/`hold`/`shot`/`wait_for*` is a narrated
hold and is silent; one spanning `run`/`click`/`goto`/… warns. The message
states what was measured and no longer asserts a cause.

## What this does

The recorder measures the encoded mp4 over the region the app occupies, and
writes the answer where the text tier is read: `content` in the `timeline.json`
envelope, plus a line on stderr as the take ends. It warns; it never refuses.

```json
"content": { "measured": true, "note": null,
             "rect": [74, 110, 1132, 432], "sample_fps": 2, "frames": 47,
             "score": 14.1, "floor": 1.0,
             "static_for": 5.0, "static_from": 13.5, "static_limit": 15.0,
             "warnings": [] }
```

Two arms, because the defect this came from defeats one of them:

* **`score`** — median luma standard deviation over the rect. Catches a
  recording that never rises above blank.
* **`static_for`** — the longest stretch where nothing inside the rect changed.
  Catches a recording that is *covered*, which the score cannot: the #91 card
  is dark with a line of white text on it and scores 12.4, comfortably
  "content".

`stitch()` carries each part's report into `segments` and reports the worst of
them on the envelope, with every warning tagged by segment. `content_report()`
is exported so a demo somebody already has can be re-checked without
re-recording.

## Acceptance: the reference demo's take 1 would have been caught

The broken take is not committed, so it was reconstructed — `record.py` from
`examples/ticket-queue/demos/2026-07-26-status-filter/` with the single line
`rec.interlude("")` removed, which is exactly what #91 was. Recorded against
the real app on this branch:

```
demo-video: WARNING — part2.seg.mp4: the content rect did not change for 23.0s,
from 0.0s. The beats in this timeline ran during that stretch and nothing in
the picture moved, so whatever they did is not visible: a title card or modal
left up over the app (issue #91), an overlay that never faded, or an app that
stopped painting. Watch part2.seg.mp4 at 0.0s before believing the beats above
it.
demo-video: WARNING — demo.mp4: segment 'part2': the content rect did not
change for 23.0s, from 0.0s. …
```

…and in the committed `timeline.json`:

```json
"content": { "measured": true, "score": 12.39, "floor": 1.0,
             "static_for": 23.0, "static_limit": 15.0,
             "warnings": ["segment 'part2': the content rect did not change for 23.0s…"] }
```

The same storyboard **with** the line restored, on the same branch, says
nothing:

```
part1.seg.mp4 shows a picture (content 15.29 over the app rect, longest still stretch 4.5s)
part2.seg.mp4 shows a picture (content 14.26 over the app rect, longest still stretch 5.0s)
demo.mp4 shows a picture (content 14.26 over the app rect, longest still stretch 5.0s)
```

Nobody opened the video for either verdict.

## Two measurements that decided the design

**The rect, not the frame.** `content-shown/demo.mp4` with the app rect painted
flat black by `drawbox` — a real recording, chrome intact, blank where the app
was:

| | healthy | app painted flat |
|---|---|---|
| over the content rect | 6.58 | **0.21** |
| over the whole frame | 82.25 | **90.33** |

The whole frame ranks the blank recording *higher*, which is #17's finding
reproduced on this fixture. Both rows are now asserted in `tests/smoke`, the
second included: if a future change scores the frame again, or the fixture
stops exhibiting the anti-correlation, the suite says so instead of passing on
a number that grades chrome.

**Counting pixels, not averaging them.** The first implementation used mean
absolute luma difference between consecutive samples, and it could not work.
libx264 re-quantises a held frame at every I-frame, which redrew the occluding
card's glyph edges hard enough to move the mean by **0.43** — while the
smallest *real* change in the healthy take of the same storyboard moved it by
**0.71**. A 1.6x gap is a coin toss. Counting reduced pixels that moved by more
than 12 luma levels separates the two phenomena by their shape instead:

| | moved pixels, per consecutive pair |
|---|---|
| covered (card over the whole terminal) | 0, on every one of 46 pairs |
| healthy terminal, real changes | 2, 13, 20, 27, 31, … 1256 |
| healthy web, real changes | 10, 13, 14, 16, 17, … 991 |

The bar sits at 4. With the mean-based bar the covered take reported 12.5s of
held picture instead of 23.0s; with this one it reports the whole segment.

## Tests

A tenth axis in `tests/smoke` (`--content-only`): **one storyboard recorded
twice**, differing only in whether `interlude("")` takes the card back down.
Same three commands, same captions, same exit codes, same `evidence/`.

| | `content-shown` | `content-covered` |
|---|---|---|
| `content.warnings` | empty | says "did not change" |
| stderr from the take (captured) | no warning | `WARNING`, unasked |
| `static_for` | under `static_limit` | ≥ 75% of the take's duration |
| five stills cropped to `#__term_host` | 5 distinct files | 1 file, five times |
| `evidence/` | ≥ 5 distinct screens | ≥ 5 distinct screens |

…plus one assertion across the pair: the video at the first and last `run` beat
must be **at least 8 dB PSNR further apart in the covered take than in the
shown one**.

Every row is a comparison, because each has a way to pass vacuously alone. The
last row is the control that makes the two above it mean anything — it is what
says the five commands really printed five different things, so "the frames
are identical" is a statement about the *recording*. It is also #97's finding
in one line: the text tier says the demo worked, the pixels say nothing was
shown.

**The primary assertions share no helper with the check.** The stills are
compared byte for byte (lossless PNGs, no threshold at all); the video is
compared with ffmpeg's PSNR; the rect the harness crops to is its own reading of
the live `#__term_host`, never the recorder's report.

One assertion does use the production scorer, and it is worth naming rather than
glossing: `_check_scored_region` imports `content_report` to grade the #17
anti-correlation. That is deliberate — the claim being made is *about* that
function's behaviour on a known-bad input — but it means that one row is not
independent of the code it grades, and an earlier version of this section
claimed otherwise.

`check_content_healthy()` also hangs off *every* graded take — web, terminal and
the stitched segments demo — because the check has to be wired into three
places and "it works on the take written to test it" is how a feature ships
broken everywhere else.

**`CONTENT_STATIC_WARN_S` is graded as a band, not trusted as a constant.** It
must sit at least 1.6x above the healthy take's longest still stretch and at
most 0.7x of the covered take's, both derived from that run's own recordings:

```
smoke: content static limit 15.0s sits in [8.8, 21.3]s (healthy 5.5s, covered 30.5s)
```

## A bar that was tuned to this box, and how CI said so

Worth recording, because it is a catalogue entry caught in the act. The first
version of the video comparison asserted an **absolute** bar: "the covered
take's two sampled moments read at least 40 dB PSNR". It measured 47.5 dB here
and CI went red at **39.8 dB** — a different x264 build re-quantising a held
frame slightly differently — on a recorder that was working perfectly.

It is now relative between the two takes, in the same run on the same encoder,
which is the pattern this suite already uses for the caption band:

| | this box | CI runner |
|---|---|---|
| covered | 47.5 dB | 39.8 dB |
| shown | 25.5 dB | 25.9 dB |
| gap | 22.0 dB | 13.9 dB |

The same run also showed the `terminal` take holding one frame for **10.0 s**
on CI against 9.5 s here, which is why `CONTENT_STATIC_WARN_S` is 15 s rather
than 12: at 12 the graded terminal take had 1.26x of headroom, and a bar with
that much margin against an unrelated storyboard's idle is a flaky red waiting
to happen.

## A limit found while building the fixture

Lengthening the content storyboard from three commands to five made the
**healthy** take report 9.5 s of held picture while it was visibly printing.
The cause is real and is now documented in `SKILL.md`: the content rect drops
the bottom fifth of the app (the caption bar is burned in there), so on a
terminal that is the last few rows *before the screen starts scrolling*. Once
it scrolls, every new line moves the whole picture and the window closes.
There is no room to shrink the trim — the caption's shadow starts at ~82% of
the app rect — so it is a stated limit, and the fixture's commands are capped
at three lines each to stay clear of it.

## Fault injection

Eleven injections, each a single exact string with the harness aborting unless
it matched exactly once, each run against the suite, each restored afterwards.
**All eleven were caught.** Full transcript:

```
--- break: any pixel difference counts as a change (a mean-style bar)
    in core.py: CONTENT_PIXEL_DELTA = 12
    exit 1, 4 failure(s)
    - content-covered: the recorder did not report a held picture for a take whose title card covered every beat. This is issue #91 reproduced, and `content.warnings` says []
    - content-covered: nothing on stderr said the recording holds still. …
    - content-covered: the card covered the take but the recorder reports only 9.0s of held picture in 31.32s of video — under 75%. …
    - content: the recorder warns at 15.0s, outside the [5.6, 6.3]s band this run measured — a healthy take of this storyboard holds 3.5s and the covered one holds 9.0s. …
    PASS

--- break: nothing short of a whole repaint counts as a change
    in core.py: CONTENT_PIXEL_DELTA = 12
    exit 1, 3 failure(s)
    - content-shown: the recorder warned about the take whose card *is* taken down: ['the content rect did not change for 31.5s, from 0.0s. …']
    - content-shown: stderr warned about a healthy take, so the warning above says nothing
    - content: the recorder warns at 15.0s, outside the [50.4, 21.7]s band this run measured — a healthy take of this storyboard holds 31.5s and the covered one holds 31.0s. …
    PASS

--- break: the still-picture limit is doubled
    in core.py: CONTENT_STATIC_WARN_S = 15.0
    exit 1, 1 failure(s)
    - content: the recorder warns at 30.0s, outside the [8.8, 21.3]s band this run measured — a healthy take of this storyboard holds 5.5s and the covered one holds 30.5s. Below the band every honest demo warns; above it, the take this check exists for does not
    PASS

--- break: the still-picture limit is halved
    in core.py: CONTENT_STATIC_WARN_S = 15.0
    exit 1, 1 failure(s)
    - content: the recorder warns at 7.5s, outside the [8.8, 21.7]s band this run measured — a healthy take of this storyboard holds 5.5s and the covered one holds 31.0s. …
    PASS

--- break: the blank floor drops to zero, so nothing is ever blank
    in core.py: CONTENT_BLANK_FLOOR = 1.0
    exit 1, 1 failure(s)
    - content: scored over the content rect, the blanked recording (0.21) and the healthy one (6.64) do not straddle the 0.0 floor — the score arm cannot tell a blank recording from a working one
    PASS

--- break: the verdict never reaches stderr
    in core.py: if not isinstance(content, dict):
    exit 1, 1 failure(s)
    - content-covered: nothing on stderr said the recording holds still. A reviewer who does not open timeline.json gets no signal at all. …
    PASS

--- break: the timeline stops carrying the picture at all
    in core.py: "content": self._content,
    exit 1, 2 failure(s)
    - content-shown: no measured `content` in timeline.json (None)
    - content-covered: no measured `content` in timeline.json (None)
    PASS

--- break: the terminal recorder cannot say where the terminal is
    in terminal.py: if self._host_box is None:
    exit 1, 2 failure(s)
    - content-shown: no measured `content` in timeline.json (this recorder does not know where the app sits in the frame, so nothing about the picture was measured — see content_report())
    - content-covered: no measured `content` in timeline.json (…)
    PASS

--- break: the web recorder scores the whole frame, chrome included
    in web.py: return content_rect(
    exit 1, 1 failure(s)
    - web: the recorder scored (0, 0, 1280, 576), which is not inside the app rect (128, 90, 1024, 576) this harness measured — it is grading its own window chrome, which scores well on a blank recording (issue #17)
    PASS

--- break: the covered take is not actually covered (the fixture's own control)
    in smoke: rec.interlude(CONTENT_CARD)
    exit 1, 6 failure(s)
    - content-covered: the recorder did not report a held picture for a take whose title card covered every beat. …
    - content-covered: nothing on stderr said the recording holds still. …
    - content-covered: the card covered the take but the recorder reports only 5.5s of held picture in 32.16s of video — under 75%. …
    - content: the recorder warns at 15.0s, outside the [8.8, 3.8]s band this run measured — a healthy take of this storyboard holds 5.5s and the covered one holds 5.5s. …
    - content-covered: the 5 stills show 5 different pictures where the terminal is, so the card did not cover this take and every assertion about occlusion here measures something else
    - content: the covered take's two sampled moments are 25.4 dB PSNR apart and the shown take's 25.5 dB — a gap of -0.1 dB, under the 8.0 dB one. Whatever the absolute numbers are on this encoder, the card is not holding the video visibly stiller than the storyboard does
    PASS

--- break: stitch() drops the segments' picture reports
    in core.py: "content": merge_content(records),
    exit 1, 1 failure(s)
    - segments: timeline.json carries no `content` — the recorder said nothing at all about whether the recording shows anything (#97)
    PASS

All injections caught.
```

The last one is a break in the **fixture**, not the product: it makes the
covered take clear its card, so the take stops being covered. It is the control
that stops every assertion above it from being satisfiable by a harness that
answers "identical" regardless — and note that it is the only injection all six
covered-side assertions fire on at once.

Three injections initially reported FAIL because my *expected* substring was
wrong, not because the assertion missed. Each is shown above re-run with the
corrected expectation and the suite's real message.

## Stated limits

Non-blocking, recorded here rather than as a round of review:

* **The bottom fifth of the app is not measured** — that is where the caption
  bar is burned in, and there is no room to shrink the exclusion (the caption's
  shadow starts at ~82% of the app rect). On a terminal that is the last few
  rows before the screen starts scrolling; measured at 9.5 s of "held" picture
  on a fixture that was visibly printing. Stated in `SKILL.md`.
* **A held stretch spanning a segment boundary is under-reported.** Each part
  measures its own `.seg.mp4`; two segments that each hold still for 8s across
  a cut are reported as 8s, not 16s. Under-reporting is the direction to prefer
  over a confident guess, and each part's own number is in `segments`.
* **The static arm cannot see a single character changing** (~0.02 mean luma at
  the 160x90 reduction, far under the bar) and does not try to. A blinking
  cursor is not the demo showing something.
* **A demo that legitimately holds one frame for more than 15s will warn.** It
  is a warning, and the sentence it prints is true; the limit is graded as a
  band rather than tuned to make that never happen.
* **The blank floor is one number for both media, and it is deliberately low
  (1.0).** The suite keeps its own higher per-medium floors as a gate. The
  shipped floor goes to somebody else's app, where a false alarm costs the
  warning its credibility everywhere — and the score arm has a partner: a page
  that never painted does not change either, so the static arm catches the
  blank case a second time.
* **The reference demo's committed `timeline.json` files predate this field.**
  They carry no `content`; the next re-record adds it. Re-recording them here
  would churn the stills for no verification gain.

## Not demoable

This change's acceptance criterion is verified by reading stderr and a JSON
field, not by watching — per `CLAUDE.md`'s test, a video of it would be
ceremony. **No demo video is expected.** The `ticket-queue demo` job does not
fire on this branch either: its `paths:` filter is
`examples/ticket-queue/**`, `.github/workflows/{ticket-queue-demo,demo-video}.yml`
and `.github/scripts/demo-*`, and this diff touches none of them.

## Verification run

* `tests/smoke` (full suite, every take): **PASSED**
* `tests/smoke --content-only`: **PASSED**
* 11 fault injections: **all caught**
* `ruff check`, `ruff format --check`, `mypy skills/demo-video/helpers/`: clean

### DIFF
--- a/skills/demo-video/README.md
+++ b/skills/demo-video/README.md
@@ -51,6 +51,14 @@ and a real terminal session (voice on):
   direction, HTML entities and JSON escaping — a stated list, not an open
   promise. It is **not committed**; `SKILL.md` explains every one of those
   decisions and where they stop.
+- **A statement about the frames, not only the storyboard** — `content` in
+  `timeline.json`, plus a line on stderr, saying whether the recording shows
+  anything at all: how much picture there is where the app sits, and the
+  longest stretch in which nothing there changed. Every other artifact here
+  describes what the storyboard *did*, and all of them can be exactly right
+  over a recording that is blank or covered — a title card over a terminal
+  once cost this project 24.3 s of a 60.2 s demo with every beat, exit code
+  and evidence file reporting success. It warns; it never fails a take.
 - **A recording that reproduces, when you ask for it** — the browser's
   timezone and locale are always pinned, and `Recorder(deterministic=True)`
   additionally freezes the page's clock and flattens animations, so the same
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -812,6 +812,12 @@ working one.
   caption bar sits, and there is no room to shrink the exclusion. On a terminal
   demo that is the last few rows *before the screen starts scrolling*. Once it
   scrolls, every new line moves the whole picture and the blind window closes.
+- **A caption long enough to wrap reaches back into the measured region.** The
+  bar grows upward, so a two- or three-line caption crosses the exclusion and a
+  caption change then does count as the picture changing. Measured at 266 pixels
+  per wrap. The effect is to make a held stretch read *shorter* than it is, and
+  it cannot touch an occluded take at all, because a card covers the caption
+  too. Keep captions to one line if you want `static_for` to mean what it says.
 - **A run of held frames is measured per segment.** Two parts of a stitched
   demo that each hold still for 8s across the cut are reported as 8s, not 16s.
 
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -721,6 +721,62 @@ key is fine, renaming one is not:
   across a merged demo. `segment` and `segment_index` (its position within its
   own segment) survive the merge untouched — use that pair, not `index`, to
   name a beat in anything that has to line up across a re-stitch.
+- `content` is the only key here that describes the **frames** rather than the
+  storyboard. See the next section.
+
+### `content` — did the recording show anything?
+
+Every other field above is a statement about what the storyboard *did*, and all
+of them can be exactly right while the recording shows nothing: a title card
+left up, a modal that never closed, an app that stopped painting. That happened
+here, on this skill's own reference demo — a card covered the terminal for 24.3s
+of a 60.2s take, and the beat table, the exit codes, the evidence files and the
+stills all read like a healthy demo.
+
+So the recorder also measures the picture, over the region the app occupies in
+the encoded frame, and writes down what it found:
+
+```json
+"content": { "measured": true, "note": null,
+             "rect": [74, 110, 1132, 432], "sample_fps": 2, "frames": 47,
+             "score": 14.1, "floor": 1.5,
+             "static_for": 5.0, "static_from": 13.5, "static_limit": 12.0,
+             "warnings": [] }
+```
+
+- `score` is the median luma standard deviation over `rect`. Under `floor` the
+  frames are featureless — a page that never painted, a take recorded black.
+- `static_for` is the longest stretch, in seconds, where **nothing inside
+  `rect` changed**, and `static_from` is when it started. Reported always; a
+  stretch at or over `static_limit` also produces a warning.
+- `warnings` is empty on a healthy take, and each entry is one sentence saying
+  what to go and look at. They are also printed on stderr as the take ends, so
+  an author who never opens this file is still told.
+- `measured` is `false`, with `note` saying why, when the check could not run.
+  `content` itself is `null` only when the take encoded no mp4.
+
+**It warns; it never fails a take.** A demo that legitimately holds one frame
+through a long narrated beat is ordinary, and refusing a take on a heuristic is
+worse than the heuristic missing one. Treat a warning as "watch the video at
+this timestamp before believing the beats", not as a verdict.
+
+**It does not judge whether the demo shows the _right_ thing** — only whether
+it captured anything at all. `rect` deliberately excludes the recorder's own
+window chrome and its caption bar: a whole-frame score is dominated by them, to
+the point where a recording with the app painted flat scores *higher* than a
+working one.
+
+On a stitched demo each part measures its own `.seg.mp4` — two segments can be
+two media with two geometries — so the envelope's `content` is the worst of the
+parts on each arm with `rect: null`, every warning tagged with the segment it
+came from, and each part's own report under `segments`.
+
+You can re-run it on a demo you already have, without re-recording:
+
+```python
+from demo_recording import content_report
+print(content_report("demos/2026-07-26-x/demo.mp4", (74, 110, 1132, 432)))
+```
 
 A merged timeline says so, and says what it was built from:
 
@@ -729,9 +785,11 @@ A merged timeline says so, and says what it was built from:
   "recorder": "Recorder",
   "segments": [
     { "segment": "part1", "media": "part1.seg.mp4", "duration": 6.6,
-      "offset": 0.0, "beats": 6, "recorder": "Recorder", "determinism": {…} },
+      "offset": 0.0, "beats": 6, "recorder": "Recorder",
+      "determinism": {…}, "content": {…} },
     { "segment": "part2", "media": "part2.seg.mp4", "duration": 8.6,
-      "offset": 6.6, "beats": 9, "recorder": "Recorder", "determinism": {…} }
+      "offset": 6.6, "beats": 9, "recorder": "Recorder",
+      "determinism": {…}, "content": {…} }
   ] }
 ```
 
@@ -1560,6 +1618,11 @@ invisible to it.
   frame is captured the moment it paints, and a published video leaks forever.
   Decide what must not appear *before* the first `goto()` — see **Redacting
   secrets**, and read what it does not cover.
+- **Reading the beat table as proof the demo showed something.** It is proof
+  the storyboard *ran*. A card left up, a modal that never closed or an app
+  that stopped painting produces a complete, correct, entirely successful
+  timeline over a recording nobody can watch. Check `content` in the same file
+  — that is the field that describes the frames.
 - **Embedding video in markdown.** Repo-relative mp4s don't play inline in
   rendered markdown, and `demo.mp4` isn't committed anyway — open the guide
   with a still and point at the re-record command instead. GitHub plays only
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -739,8 +739,8 @@ the encoded frame, and writes down what it found:
 ```json
 "content": { "measured": true, "note": null,
              "rect": [74, 110, 1132, 432], "sample_fps": 2, "frames": 47,
-             "score": 14.1, "floor": 1.5,
-             "static_for": 5.0, "static_from": 13.5, "static_limit": 12.0,
+             "score": 14.1, "floor": 1.0,
+             "static_for": 5.0, "static_from": 13.5, "static_limit": 15.0,
              "warnings": [] }
 ```
 
@@ -765,6 +765,20 @@ it captured anything at all. `rect` deliberately excludes the recorder's own
 window chrome and its caption bar: a whole-frame score is dominated by them, to
 the point where a recording with the app painted flat scores *higher* than a
 working one.
+
+**Two limits worth knowing before you rely on it:**
+
+- **The bottom fifth of the app is not measured** — that is where the caption
+  bar is burned in, and there is no room to shrink the exclusion. On a terminal
+  demo that is the last few rows *before the screen starts scrolling*: output
+  landing there does not count as the picture changing, so a stretch that is
+  visibly printing can read as held. Once the screen scrolls, every new line
+  moves the whole picture and the blind window closes.
+- **A run of held frames is measured per segment.** Two parts of a stitched
+  demo that each hold still for 8s across the cut are reported as 8s, not 16s.
+
+Neither can produce a false *success*: both make the check quieter than the
+truth, never louder.
 
 On a stitched demo each part measures its own `.seg.mp4` — two segments can be
 two media with two geometries — so the envelope's `content` is the worst of the
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -801,7 +801,17 @@ window chrome and its caption bar: a whole-frame score is dominated by them, to
 the point where a recording with the app painted flat scores *higher* than a
 working one.
 
-**Three limits worth knowing before you rely on it:**
+**Five limits worth knowing before you rely on it:**
+
+- **A held stretch containing only narration is never reported.** This is the
+  verb correlation doing its job, and it is also the one blind spot that can
+  hide a real fault: a demo that raises an interlude card and then only
+  captions, holds and takes stills behind it is silent, however long the card
+  stays up. Measured: a card over 31.5s of a 34s take, no warning. There is no
+  third answer available from the frames — with the caption band excluded, an
+  honest tour of a still screen and a card left up are byte-identical. If your
+  storyboard raises a card, take it down explicitly; do not rely on this check
+  to notice.
 
 - **A caption change is invisible to the held-picture arm, by design.** The
   caption bar is the recorder's own drawing and it renders over an interlude
@@ -815,17 +825,29 @@ working one.
 - **A caption long enough to wrap reaches back into the measured region.** The
   bar grows upward, so a two- or three-line caption crosses the exclusion and a
   caption change then does count as the picture changing. Measured at 266 pixels
-  per wrap. The effect is to make a held stretch read *shorter* than it is, and
-  it cannot touch an occluded take at all, because a card covers the caption
-  too. Keep captions to one line if you want `static_for` to mean what it says.
+  per wrap. The effect is to make a held stretch read *shorter* than it is. It
+  cannot touch a take occluded by the recorder's **own** interlude card, which
+  is opaque and paints above the caption — but an app-level overlay or modal is
+  app DOM, the caption paints *over* it, so a wrapping caption can split a real
+  occlusion into stretches that each sit under the limit. Measured: 25s becomes
+  11s + 8s + 6s, and goes silent. Keep captions to one line if you want
+  `static_for` to mean what it says.
 - **A run of held frames is measured per segment.** Two parts of a stitched
   demo that each hold still for 8s across the cut are reported as 8s, not 16s.
 
-None of the three can produce a false *success*: each makes the check quieter
-than the truth, never louder.
+Three of the five can hide a real occlusion — the first, the fourth and the
+fifth. A card behind narration only, a wrapping caption splitting an app-level
+overlay into short stretches, and a hold that straddles a segment cut each
+produce a clean report on a take that shows nothing.
 
-Neither can produce a false *success*: both make the check quieter than the
-truth, never louder.
+The second and third cannot: excluding the caption bar, and not measuring the
+bottom fifth, both mean *less* of the picture counts as moving, so a stretch
+grows rather than shrinks. They push toward reporting a hold that is not there,
+not toward missing one.
+
+**A silent content report is not a guarantee that the demo shows your app.** It
+rules out the two loud failures — a recording that never rises above blank, and
+a screen that never moved while a verb was acting on it — and nothing more.
 
 On a stitched demo each part measures its own `.seg.mp4` — two segments can be
 two media with two geometries — so the envelope's `content` is the worst of the
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -740,25 +740,60 @@ the encoded frame, and writes down what it found:
 "content": { "measured": true, "note": null,
              "rect": [74, 110, 1132, 432], "sample_fps": 2, "frames": 47,
              "score": 14.1, "floor": 1.0,
-             "static_for": 5.0, "static_from": 13.5, "static_limit": 15.0,
+             "static_for": 21.5, "static_from": 3.5, "static_limit": 15.0,
+             "static_beats": [{ "index": 6, "verb": "caption", "acting": false },
+                              { "index": 7, "verb": "hold", "acting": false }],
              "warnings": [] }
 ```
 
 - `score` is the median luma standard deviation over `rect`. Under `floor` the
   frames are featureless — a page that never painted, a take recorded black.
 - `static_for` is the longest stretch, in seconds, where **nothing inside
-  `rect` changed**, and `static_from` is when it started. Reported always; a
-  stretch at or over `static_limit` also produces a warning.
+  `rect` changed**, and `static_from` is when it started. Both are reported
+  always, whatever they are.
+- `static_beats` is what the storyboard was doing during that stretch, and it
+  is what decides whether the stretch is worth a warning. See below.
 - `warnings` is empty on a healthy take, and each entry is one sentence saying
   what to go and look at. They are also printed on stderr as the take ends, so
   an author who never opens this file is still told.
 - `measured` is `false`, with `note` saying why, when the check could not run.
   `content` itself is `null` only when the take encoded no mp4.
 
-**It warns; it never fails a take.** A demo that legitimately holds one frame
-through a long narrated beat is ordinary, and refusing a take on a heuristic is
-worse than the heuristic missing one. Treat a warning as "watch the video at
-this timestamp before believing the beats", not as a verdict.
+### A long held stretch is not a problem, and this is the part to understand
+
+The example above holds one picture for **21.5s** — well past `static_limit` —
+and produces no warning at all. That is correct, and it is not a special case:
+
+> `static_for` on its own cannot tell a healthy demo from a broken one.
+
+Measured on real takes: a demo touring a rendered screen with three captions
+holds the measured region for **20–22s**. The title card that covered this
+skill's own reference demo held it for **23s**. There is no threshold between
+those two numbers, because the region measured excludes the recorder's caption
+bar — and swapping captions over a still screen is exactly what this guide tells
+you to do during a wait.
+
+So the warning needs a second condition: **did a verb that acts on the app run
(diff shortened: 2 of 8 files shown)


rogvid:
Follow-up commit `ce9e55a` adds one bullet to `skills/demo-video/README.md`'s "What you get" list, so the first thing a reader sees about this skill mentions the field that describes the frames. No code change; CI re-ran on the new head.

rogvid:
CI is green on `664672c` — all six checks, including `smoke` (9m31s).

The two bars the previous run exposed now have room on the runner:

```
smoke: terminal content ok (score 8.44, longest still 10.0s under 15.0s, …)
smoke: content static limit 15.0s sits in [9.6, 21.3]s (healthy 6.0s, covered 30.5s)
smoke: content video held 46.8 dB against moved 25.5 dB — 21.3 dB apart, over the 8.0 dB gap
smoke: content rect 6.58 vs blanked 0.21 (floor 1.0); whole frame 82.28 vs blanked 90.39 — the frame ranks the blank recording higher, the rect does not
```

Note the PSNR gap read 21.3 dB here against 13.9 dB on the run that failed the old absolute bar, and 22.2 dB locally — which is the spread the relative comparison exists to absorb. The whole-frame anti-correlation reproduces on the runner too (90.39 blank vs 82.28 healthy).

rogvid:
Blocking finding fixed in `052dbc3`. You were right on all of it, including that the trim is not the lever.

## The fix

The stretch is now correlated with the beat log, which the recorder has in memory at the point it writes the report. A held picture spanning only `caption`/`interlude`/`bridge`/`hold`/`pause`/`shot`/`wait_for*` is a **narrated hold** and is silent; one spanning `run`/`send`/`key`/`click`/`type_into`/`goto`/`scroll_to`/`spotlight`/`move_to`/`redact` warns. `content.static_beats` carries the classified list, so a reader can see *why* it did or did not fire.

`wait_for*` are passive deliberately: a wait that returns immediately because its condition already held changes nothing, and cannot be told apart here from one that really waited. Guessing would put the false positive back.

**Verified on your reproductions**, recorded against the real ticket-queue app:

```
terminal, 3 touring captions:
  demo.mp4 shows a picture (content 14.13 over the app rect, longest still
  stretch 21.5s — over the 15.0s limit, but every beat inside it was narration,
  a hold or a wait, which is what a still screen is supposed to look like)
```

And to isolate the mechanism rather than the constant — re-running `content_report()` over both healthy takes with `static_limit` **forced below their own held stretch**:

```
== web       held 13.0s, limit forced to 11.0s
   warnings=0   spanned: wait_for, caption, hold, shot, caption, hold  (all passive)
== terminal  held 21.5s, limit forced to 19.5s
   warnings=0   spanned: wait_for_prompt, shot, caption, hold, caption, hold, caption, hold
```

Silent because of the correlation, not because of the threshold.

The reconstructed reference take 1 still fires, and now names what it caught:

```
demo-video: WARNING — part2.seg.mp4: the content rect held one picture for 21.5s
(0.5s-22.0s) while 3 verb(s) that act on the app ran inside it: beat 2 `run`,
beat 7 `run`, beat 12 `run`. Nothing changed in the measured region while they
ran. **What was measured**: the app rect (74, 110, 1132, 432), which excludes
the recorder's own caption bar — so a caption change is invisible here by design
and is not what this reports. An overlay left up, a modal that never closed and
an app that stopped painting all look like this; so does a demo that genuinely
holds still through these verbs. The frames cannot tell those apart — open
part2.seg.mp4 at 0.5s and look.
```

No asserted cause; it names the region it measured and says outright that the frames cannot discriminate. The two banned phrasings are now **asserted against** in the suite, so the message cannot drift back.

## The fixture can now reach the false-positive path

`content-toured/` — a third take, one command then a rendered screen narrated with touring captions, written the way SKILL.md tells authors to. Asserted **silent while holding past the limit**:

```
smoke: content-toured held one picture for 25.0s, over the 15.0s limit, across
['wait_for_prompt', 'shot', 'caption', 'hold', 'caption', 'hold', 'caption', 'hold']
— and the recorder said nothing, which is correct
```

Three assertions, because "no warning" alone would also pass on a recorder whose static arm was switched off:

1. `static_for >= static_limit` — the premise. Without it the take would keep passing after the thing it grades regressed.
2. `warnings == []` and nothing on stderr.
3. `static_beats` non-empty **and containing no acting verb** — the mechanism, not the symptom.

Your point about *why the band could not catch this* is recorded in `tests/README.md` and in the `CONTENT_COMMANDS` comment, named as the config-hidden path: a fixture shaped to avoid a failure mode cannot grade it.

I also added `check_verb_classification()`, which scrapes `@_beat_verb("…")` and `self._beat("…")` out of the package source — scraped rather than imported, so an omission cannot hide on both sides — and requires each of the 22 verbs to be in exactly one set. An unclassified verb silently counts as passive, which would switch the arm off for a whole class of demo.

## The docs said the wrong cause

Rewritten. SKILL.md now leads with the table above under *"A long held stretch is not a problem"*, states that `static_for` alone cannot separate healthy from broken, and lists three limits instead of the one wrong one — the caption-band blindness first, the terminal pre-scroll rows second, per-segment third, with the note that none can produce a false success. There is also a new *Common mistakes* entry: **"Reading `content.static_for` as a score. It is not one."**

## Cheap fixes

- **"The check and its test share no helper"** — corrected in the PR body. It now says the *primary* assertions are independent and names `_check_scored_region` as the one that deliberately imports `content_report` to grade the #17 anti-correlation, since the claim is about that function's behaviour on known-bad input.
- **`content_report()`'s "Never raises"** — now true. The whole body is in a guarded wrapper (`_content_report` does the work), and the docstring says why it matters: it runs in `__exit__` before the timeline, evidence and frames are written. `content` is also scrubbed with the rest of the document now — it carries no selector today, and I kept it that way deliberately (verb and index only, since a selector can hold a registered secret and this string reaches stderr before scrubbing), but the field should not be the one place the mask does not reach.

## Issues filed

- **#121** — `merge_content`'s "worst" direction is ungraded. Behaviour verified correct; flipping `max`↔`min` passes the suite because both segments of the `segments` take are healthy.
- **#122** — the PSNR gap rides one encoder-sensitive number (`covered` 47.5/39.8, `shown` 25.5/25.9, gap 22.0→13.9 against an 8.0 bar).

## Verification

- `tests/smoke` full suite: **PASSED**
- **16 fault injections, all caught.** Five are new and target this fix directly:

```
--- break: the held-picture arm ignores the beat log and warns on duration alone
    in core.py: if held >= static_limit and acting:
    - content-toured: the recorder warned about a healthy demo that narrates a rendered
      screen — nothing is occluded, the terminal is fully drawn, and every beat in the
      held stretch is a caption or a hold. …"while 0 verb(s) that act on the app ran
      inside it: "…
    - content-toured: stderr carried a held-picture warning for a healthy take
    PASS

--- break: every verb counts as acting, so a narrated hold warns
    in core.py: "acting": verb in CONTENT_ACTING_VERBS,
    - content-toured: the held stretch is reported as spanning acting verbs
      ['wait_for_prompt', 'shot', 'caption', 'hold', …], but this storyboard touches
      nothing after its one command
    PASS

--- break: the warning goes back to naming a cause it cannot know
    in core.py: f"app that stopped painting all look like this; so does a demo that "
    - content-covered: the warning asserts 'title card or modal left up', a cause the
      frames cannot establish — a held stretch has honest explanations too (see
      content-toured)
    PASS

--- break: a new verb is added and classified as neither
    in core.py: "wait_for_text",
    - content: ['wait_for_text'] record beats but are in neither CONTENT_ACTING_VERBS
      nor CONTENT_PASSIVE_VERBS. An unclassified verb counts as passive, so a held
      picture spanning it never warns
    PASS

--- break: the beat log never reaches the check
    in core.py: self._content = content_report(self._media_path(), rect, self._beats)
    - content-covered: the recorder did not report a held picture for a take whose title
      card covered every beat
    - content-toured: content.static_beats is None. The take held a picture for 24.5s and
      the recorder recorded no beats inside it, so 'no acting verb ran' is vacuous rather
      than true
    PASS
```

One injection **aborted** rather than reporting a pass — `"content": self._content,` no longer matched after I added scrubbing to that line. The harness refused to proceed, which is what it is for; retargeted and re-run.

Not merging.

rogvid:
Follow-up in `0f7c593`. CI failed the toured take's **own premise assertion** — `static_for >= static_limit` — which is precisely why that assertion exists. Two real things behind it.

## A wrapping caption reaches back into the measured region

The touring captions were sentence-length. On the runner two of them wrapped to three lines, and **a caption bar that wraps grows upward** — past the bottom fifth the content rect trims off and into the measured region. 266 changed pixels per wrap, chopping the held stretch from 25.0s to 11.0s:

```
== CI/content-toured    static 11.0s from 18.0s
    3.5s moved 8      4.0s moved 37     11.0s moved 266    18.0s moved 266
== local/content-toured static 25.0s from 4.0s
    3.5s moved 8      4.0s moved 38
```

The fixture now uses captions short enough that they cannot wrap, with explicit holds. The behaviour is a stated limit in `SKILL.md` now, and it is a **safe** one: it makes a stretch read *shorter* than it is, never longer, and it cannot reach an occluded take at all, because a card covers the caption too.

## A beat's midpoint was the wrong test for "inside the stretch"

With shorter captions the take then warned — correctly by the old rule, and wrong in fact:

```
run("seq 1 4")   2.32s – 2.85s     midpoint 2.585s
output lands     2.50s
held stretch     2.50s – 31.50s
```

By midpoint that `run` is "inside a stretch where nothing changed", when it is the thing that **ended the previous stretch and began this one**. That is the same false-positive shape you blocked on, one layer down.

A beat now qualifies only if it **began** at or after the stretch did. What that trades away is a long acting verb that starts before a stretch and runs silently through it — now missed. Quieter than the truth, never louder, and it is also the direction the beat/video clock skew (#18) pushes an edge case.

## Now

```
smoke: content-toured held one picture for 29.0s, over the 15.0s limit, across
['wait_for_prompt', 'shot', 'caption', 'hold', 'caption', 'hold', 'caption', 'hold']
— and the recorder said nothing, which is correct
```

Reconstructed reference take 1 still warns at 23.0s naming its three `run()` beats; your touring terminal take (21.5s) and web take stay silent. Full suite green locally.

**17 fault injections, all caught**, including a new one for the boundary rule:

```
--- break: a beat that ended the previous stretch counts as inside the next one
    in core.py: if not start <= float(t_start) <= end:
    exit 1, 3 failure(s)
    - content-toured: the recorder warned about a healthy demo that narrates a
      rendered screen … "while 1 verb(s) that act on the app ran inside it: beat 1 `run`"
    - content-toured: stderr carried a held-picture warning for a healthy take
    - content-toured: the held stretch is reported as spanning acting verbs ['run'],
      but this storyboard touches nothing after its one command
    PASS
```

rogvid:
CI green on `0f7c593` — all six checks, `smoke` in 10m9s. The runner reproduces the toured take at the same 29.0s it holds locally, and says nothing about it:

```
smoke: all 22 storyboard verbs are classified (14 acting, 8 passive)
smoke: content static limit 15.0s sits in [9.6, 21.3]s (healthy 6.0s, covered 30.5s)
smoke: content-toured held one picture for 29.0s, over the 15.0s limit, across
  ['wait_for_prompt', 'shot', 'caption', 'hold', 'caption', 'hold', 'caption', 'hold']
  — and the recorder said nothing, which is correct
smoke: terminal content ok (score 8.44, longest still 10.0s against a 15.0s limit, …)
smoke: content video held 46.8 dB against moved 25.4 dB — 21.4 dB apart, over the 8.0 dB gap
smoke: content rect 6.57 vs blanked 0.22 (floor 1.0); whole frame 82.29 vs blanked 90.38
```

Worth noting the graded `terminal` take reads 10.0s on the runner against 9.5s here and no longer matters either way — the assertion that it stay under the limit is gone, because a healthy take owes silence rather than a small number. Not merging.

PR #125, opened 2026-07-27
Open a web take on the app, not on 400ms of blank white

### LINKED ISSUE #119: A web take opens on ~400ms of blank white before the app paints
Found by a human watching the CI-recorded empty-state demo (#117). The web twin
of #110, which was fixed for the terminal in #113.

## Measured

App rect (inside the window chrome), first second of `empty-state-demo.mp4`:

| t | distinct byte values in the app rect |
|---|---|
| 0.00 | 3 — flat white |
| 0.16 | 3 |
| 0.32 | 5 |
| 0.48 | **228** — painted |

So roughly **400ms of blank white** opens every web take.

## Cause

`Recorder._start()` renders the window chrome and background **once, to a static
PNG** (`.frame.png`) on a throwaway page; `_postprocess` composites the recording
into it. Playwright's screencast therefore records only the *app page* — and that
page is `about:blank` until `goto()` completes. `about:blank` paints white, so
the composited result is a white rectangle sitting inside a correct-looking
window frame, which reads as "the app loaded blank".

## Correcting the record

#110's analysis stated the web recorder was immune: *"Recorder.__enter__ has
nothing on screen until goto() — there is no 'before' to flash."* That is wrong.
There is a before, and it is white. The terminal case was more obvious only
because an empty terminal with a prompt looks broken, whereas a white rectangle
looks like a page still loading — which is the more dangerous version, because it
reads as plausible.

## Directions

- **Trim the leading blank frames in `_postprocess`.** Detect the first frame
  whose app rect is not uniform and cut before it. Cheap and reliable. **But it
  shifts every beat timestamp**, so the trimmed offset has to be recorded and
  applied the same way capture loss is (#18), or `frames/` extraction (#8) and
  the merged offsets in `stitch()` (#7) all land early. That interaction is the
  whole cost of this option and should be settled before anyone starts.
- **Hold the first frame** — do not begin composing output until the app page has
  painted once. Cleaner semantically; Playwright starts the screencast with the
  context, so this is still a post-hoc trim, just expressed differently.
- **Paint the gap deliberately.** Give the app page a background matching the
  window chrome so the opening reads as "window not yet open" rather than "app
  loaded blank". Cosmetic, cheapest, and does not touch any timestamp — a
  reasonable interim if the trim's ordering cost is judged too high.

## Acceptance

A web take's first frame shows the application, or shows something deliberate.
No flat-white app rect at t=0. If timestamps are shifted to achieve it, the shift
is recorded in `timeline.json` and `frames/` still lands on the right beats.

## Related

#110 / #113 (the terminal half, fixed), #18 (the existing capture-loss offset
this must not fight with), #97 (nothing verifies the recording shows anything —
a content check would have flagged 400ms of blank automatically).

rogvid:
Fixes #119.

## Acceptance criterion

A web take's first frame shows the application. No flat-white app rect at t=0,
and if timestamps are shifted to achieve it, the shift is recorded and `frames/`
still lands on the right beats.

**Measured on `examples/ticket-queue/demos/2026-07-27-empty-state` — the same
storyboard whose CI recording raised the issue.** Distinct luma values in the
app rect:

| t | before (#119) | after |
|---|---:|---:|
| 0.00 | 3 — flat white | **216** |
| 0.16 | 3 | 216 |
| 0.32 | 5 | 216 |
| 0.48 | 228 | 217 |

No timestamp moved: `held` is an overlay switched off part way through, so the
duration, the audio and every beat time are byte-for-byte what they were.

## What it does, and why not the options in the issue

Chromium's screencast starts with the page, and the page is `about:blank` until
the first `goto()` returns. `_t0` cannot move later to skip it — the comment on
it in `__enter__` is load-bearing and says why.

**Trimming** is the honest fix and costs the most: a uniform `t -= trim` across
beats, issues, `frames/` extraction and `stitch()`'s merged offsets, colliding
with the capture-loss offset in #18. **Painting the gap** costs nothing and does
not give you the app at second 0.

So the gap is **covered, not cut**: `_postprocess` finds the first frame that
differs from the one the take opened on and composites that frame over the app
rect for the seconds before it, as a second overlay with
`enable='lt(t,held)'`. Content at every later time keeps the timestamp it
already had.

## Stated limits

- **The opening frames show the app before it painted.** That is a picture the
  recording did not capture. `content.opening.held` carries it, the stderr
  summary says it on every take that has one, and SKILL.md has a section. A
  video that repairs itself silently is one nobody can audit.
- **The video is not a measurement of load time**, and never was a good one —
  the gap it covers is the recorder's own startup as much as the app's.
- Past `OPENING_HOLD_LIMIT_S` (1.5s) nothing is held and the take warns instead.
- **Web only.** The terminal recorder frames itself in the page; `held` is null
  there, and the measured `gap` is reported but never warns, because a warning
  that fires on every take of a medium stops being read. Whether a terminal take
  opens blank is not graded here.
- Only the `web/` take is asserted. The web segment inside `segments/` and the
  entropy takes hold their openings too, and nothing checks it there.

## The number that grades it is measured from the other side

`held` is what the recorder believes it did. `gap` is measured **afterwards**,
off the encoded mp4, by the same sampling the picture check uses — so on a
healthy take it reads `0.0` *because* the hold landed, and a hold that silently
did nothing shows up as a non-zero `gap` on the file somebody watches, plus a
warning. Two passes over two files, on purpose.

That is not theoretical: extracting the held frame uses `-ss` before `-i`, and
if that had fast-seeked to keyframe zero it would have held the *blank* frame.
`gap` would have read 0.40 and warned. It reads 0.0.

## Fault injection

`tests/smoke --web-only` against three broken copies. Each injection aborts the
run unless its pattern matches **exactly once**.

```
--- break: the hold is composited but switched off, so nothing reaches the frame
    PASS (exit 1)
      | - web: demo.mp4 opens on a picture that is not there — its first frame
        scores 0.00 luma stddev over (128, 90, 1024, 460), under 8.00 (0.5 of
        this take's own 16.00 median). The recorder says it held the opening
        for 0.40s; whatever it composited, the app is not visible at t=0.

--- break: the opening is never detected, so no gap is ever covered
    PASS (exit 1)
      | - web: this take reports no opening hold (content.opening.held is 0.0),
        so the app had already painted when the recording started and nothing
        here graded the fix for #119.

--- break: the blank floor is removed, so a static painted picture reads as a gap
    PASS (exit 1)
      | - opening-gap: a video that shows one *painted* picture for its whole
        length measured a gap of 1.95, not 0.0.

All 3 injections were caught.
```

### The injection found a defect in the assertion itself

The first version of `check_opening` sampled with `sample_fps=1` and graded
`frames[0]`. The `fps` filter quantises to its own output slots, and at 1 fps
slot zero is a whole second wide — measured, it returns a frame from around
0.5s, which on a take with a 0.4s blank opening is **after** the blank has
ended.

So it scored **17.06 on a video whose first frame was flat white**, and passed.
Green on healthy and green on broken. The injection that switches the hold off
went unnoticed by the only assertion written to catch it, and was reported only
by `check_healthy` noticing the recorder's own warning — which is the recorder
grading itself.

Fixed in a0a2eb4 by cutting the window with `-t` instead: 17.25 healthy against
0.00 blanked. Reading the assertion would not have found this.

### Why the synthetic pair exists

A web take always opens blank, so **no recorded take in the suite can reach the
blank floor** — the constant that stops `opening_gap` firing on an app that
painted immediately and then held still. A fixture that cannot reach a path does
not grade the constant guarding it. `check_opening_gap` synthesises both shapes
with ffmpeg, no browser: white until 0.6s then colour bars (must measure 0.6),
and colour bars throughout (must measure 0.0). Colour bars rather than
`testsrc2` because `testsrc2` animates, and a video that changes every frame
cannot express "painted and holding still" — the distinction being graded.

## Demo

Demoable under `AGENTS.md` (verified by watching), but this branch changes the
recorder rather than `examples/ticket-queue`, so `ticket-queue-demo.yml`'s
`paths:` filter correctly does not fire. The demo was recorded locally off this
branch, from the unchanged empty-state storyboard, and handed over directly.


rogvid:
## Two stated limits, found reviewing this against the catalogue

Neither blocks; both are recorded rather than implied by silence.

**1. The refuse-to-hold branch is ungraded — filed as #126.**
`OPENING_HOLD_LIMIT_S = 1.5` decides whether the recorder covers a blank opening
or declines and warns. Nothing exercises the declining branch: the fixture app
paints in 0.30–0.50s, so every recorded take takes the cover path. Set the
constant to 100 and the whole suite still passes.

It is warn-only and fails safe — over the limit the recorder does nothing to the
picture, and the `gap` measured afterwards warns on its own. There is no
artifact it can make lie; the failure mode is a demo that opens blank, which is
the state before this branch.

**2. The audio path through the new filter chain is not exercised.**
`speech=True` needs an `ELEVENLABS_API_KEY`, which CI does not have, so every
take here encodes without an audio stream. The mapping is unchanged from before
(`-map 0:a? -c:a copy`, input 0 either way) and the hold adds only video
stages — but "unchanged and therefore fine" is an argument, not a measurement,
and it is worth saying which one this is.

## What the runner measured

Worth reading off the CI smoke log rather than my laptop, because the limit
above is the one number that could differ under contention: if a GitHub runner's
web take opened on more than 1.5s, the recorder would decline to hold and the
take would warn — turning a healthy CI run red. The holds this run reports are
the evidence for whether 1.5s has headroom on a 2-core box also encoding video.

rogvid:
## What the runner measured — all checks green

The number I said was worth reading off CI rather than my laptop, because it is
the one that could differ under contention. Every web take in the run, on a
2-core GitHub runner also encoding video:

| take | hold |
|---|---:|
| `web`, `segments` (both parts), `redaction`, `entropy` (×3), `evidence` | 0.45–0.55s |
| `web-problems` | **0.70s** — the longest in the run |

Against `OPENING_HOLD_LIMIT_S = 1.5`, that is **2.1× headroom at the worst
observed value**, up from 0.30–0.50s (3–5×) on my machine. The runner is
consistently slower, as expected.

**Stated plainly: 2.1× is a comfortable-looking margin, not an order of
magnitude**, and the catalogue in `wip/verified-review/SKILL.md` is explicit
that the first kind is worth less than the second. What sits on the other side
of it is bounded, which is why this is recorded rather than acted on: exceeding
the limit produces a warning and a blank opening — the state before this branch
— not a corrupt artifact. The failure mode is a red CI run, not a demo that
lies.

I am **not** retuning the constant on the strength of one CI run. Raising it to
buy margin also raises how much fabricated picture a take may contain, which is
the thing the limit exists to cap; moving it in either direction on today's
numbers would be tuning a constant to a box. #126 is where it gets revisited,
with a fixture that can actually reach the branch.

`smoke: web opens on the app (frame 0 scores 19.37 against a 17.63 median, over
a 0.45s hold)` — on the runner, with the same assertion that reads 0.00 when the
hold is switched off.

Issue #64, opened 2026-07-26
Run the skill on its own use case: a reference PR demonstrating the full review loop


rogvid:
Filed out of the triage in #62. **This is larger than anything currently on
the `blocks-v1` list, and the tracker did not know about it.**

## The gap

demo-video has never been run as its own use case. Everything in this repo
tests the *recorder* — `tests/smoke` records the fixture app and grades the
output. Nothing tests the **loop**: feature branch → storyboard → recording →
artifacts on a PR → a human approves in two minutes and an agent checks
conformance without decoding video.

So the central claim in #62 is unverified. We have a recorder with a test
suite and a hypothesis about how it changes review.

## What to build

A small, real, deliberately unremarkable application, and a pull request
against it that carries the full set of artifacts.

- [ ] **The app.** Something with genuine UI state to demonstrate and a
      terminal side worth showing — a CLI plus a small web front end, so both
      recorders are exercised. It must be boring: the demo is the subject, not
      the app.
- [ ] **A feature with written acceptance criteria**, phrased the way a real
      ticket is. Three to five criteria, at least one of which the
      implementation deliberately *does not* satisfy.
- [ ] **A storyboard an agent wrote** from those criteria, not one hand-tuned
      until it looked good. If an agent cannot produce a usable storyboard
      from a ticket, that is the finding.
- [ ] **The PR**, with the video reaching the reviewer by whatever #2 settles
      on, plus `timeline.md`, beat frames, and per-beat evidence.
- [ ] **Both review passes run for real** — the comprehension reviewer and,
      once #12 lands, the conformance reviewer — with their output in the PR
      thread. The undemonstrated criterion should be caught by the second one.

## What it is for

Three things, in order of how much they matter:

1. **Falsifying the premise.** The 15-60 second budget, the claim that beat
   frames plus evidence are enough for an agent, the claim that a lead
   developer can approve from a video — none of these has been tested against
   a real feature. This is where they get tested.
2. **The performance question.** #63 group 3 closed three cost issues (#39,
   #44, #49) as unmeasurable on a 60-element fixture. A real app answers them
   with a profile. If a 200-beat take is unusable, we want to know here.
3. **Adoption.** "Read SKILL.md" is a much weaker pitch than "here is a PR,
   this is what your reviewer sees." A team standardising on this will ask for
   exactly this artifact.

## Acceptance

Someone who has never seen demo-video can open the reference PR and, without
running anything, say what the feature does, which acceptance criterion is
[...]

Issue #24, opened 2026-07-25
A verb that raises logs a beat indistinguishable from one that succeeded


rogvid:
Found in review of #6.

`_DemoBase._beat` closes the record in a `finally`, so a verb that raises
still gets a complete-looking beat with a plausible `t_start` and `t_end`.
Today that is invisible in practice, because an exception aborts the take and
no timeline is written at all.

It stops being invisible as soon as anything catches. `tests/smoke` already
collects rather than raises so that a failing take still produces an mp4, and
#11 proposes dumping artifacts when a take crashes — at which point a timeline
gets written containing beats for verbs that did not do what they say.

The slice this actually matters to is #12, which turns the manifest into a
**conformance gate**: 'this beat demonstrated AC-3' must not be derivable from
a beat whose verb threw. A review gate that can be lied to is worse than no
gate.

## Fix

Record the outcome on the beat — an `error` key carrying the exception type
and message, absent on success. Additive, and it is the same seam #3 wants for
attributing console errors to the beat that was active.

## Acceptance

A beat whose verb raised is distinguishable from one that returned, by reading
`timeline.json` alone.

rogvid:
**blocks-v1** (#62), category: the artifact must never lie. Also a hard dependency of #12 (a conformance gate that can be lied to is worse than no gate) and of #11 step 1 (writing a timeline on the exception path ships beats for verbs that threw, unless this lands first).

Issue #126, opened 2026-07-27
The web recorder's refuse-to-hold branch is ungraded: no fixture paints slowly enough to reach it


rogvid:
Found reviewing my own change in #125, against the "ungraded constant" entry in
`wip/verified-review/SKILL.md`.

## The gap

`OPENING_HOLD_LIMIT_S = 1.5` decides whether the web recorder covers a blank
opening or declines and warns instead. **Nothing exercises the declining
branch.**

Measured: the fixture app paints in 0.30-0.50s on this machine, so every
recorded take in the suite takes the *cover* path. Set the constant to 100 and
the whole suite still passes. Set it to 0.01 and what fires is the premise arm
of `check_opening` ("this take reports no opening hold"), which is a different
statement — it says the take had nothing to cover, not that the recorder
correctly refused to cover too much.

This is a config-hidden path in the sense the catalogue means: the fixture
cannot reach it, so the constant guarding it is graded by nothing.

## Why it was not blocking for #125

The branch is warn-only and fails safe. Over the limit the recorder does
*nothing* to the picture and says so on stderr, and the `gap` measured
afterwards off the encoded file then warns on its own. There is no artifact it
can make lie — the failure mode is a demo that opens blank, which is exactly the
state before #119.

## What would grade it

A fixture page that defers its first paint past the limit — a query parameter on
`tests/fixture` the storyboard can ask for — recorded as one take, asserting:

- `content.opening.held` is `0.0` and `note` says why,
- `content.warnings` names the opening,
- and the frames really are blank at t=0, so the take is honest about having
  declined rather than quietly having failed.

Deliberately not included: any change to the limit's *value*. 1.5s is a
judgement about when a slow paint stops being a recorder artefact and starts
being information about the app, and this issue is about grading the branch, not
retuning the number.

## Related

#119 (the fix), #125 (the change), #61 (smoke's length — this adds a take, and
that is the cost to weigh).

rogvid:
CI measurement from #125, worth having here when this is picked up: on a 2-core GitHub runner the web takes hold 0.45-0.55s, with one at **0.70s** (`web-problems`). Against the 1.5s limit that is 2.1x headroom at the worst observed value — so the slow-paint fixture this issue asks for needs to defer paint past 1.5s deliberately, and cannot rely on runner contention to get there.

rogvid:
**This issue now closes two things, and the second one is measured.**

The original scope was the *refuse-to-hold* branch being ungraded. The CI runs on
[...]

PR #127, opened 2026-07-27
Record a take against a ticket's acceptance criteria

### LINKED ISSUE #12: Acceptance-criterion tagging and a conformance reviewer

rogvid:
Fixes #12.

## Acceptance criterion

> A storyboard tagged against a ticket's acceptance criteria produces a coverage
> report naming any criterion with no beat demonstrating it.

Recorded against `examples/ticket-queue` with three criteria, two demonstrated:

```
| criterion                                          | claimed by | at   | still                    |
| **AC-1** — The support queue lists every open …    | beat 1     | 1.11 |                          |
|                                                    | beat 2     | 3.43 | `images/01-queue.png`    |
| **AC-2** — The queue can be filtered to one status.| beat 3     | 3.48 |                          |
|                                                    | beat 4     | 5.46 | `images/02-filtered.png` |
| **AC-3** — The CLI prints the same filtered list.  | *nothing claims this* |  |             |

**1 of 3 criteria have no beat claiming them: `AC-3`.**
```

## The design decision worth arguing with

**The report records what the storyboard *claimed*, never what it proved**, and
every name in it says so — `claimed`/`unclaimed`, not `demonstrated`/`missing`.

An `ac=` tag is a string its author typed. The failure this feature exists to
catch is the **tautology** the issue itself names: a storyboard derived from the
diff rather than from the ticket produces a polished, convincing demo of a
misread requirement. A file that said *"AC-3 demonstrated"* would hand the
conformance reviewer the author's word for the exact thing it was convened to
check — a check sharing the blind spot of the bug it exists to find.

So `unclaimed` is the only machine-checkable finding here, and it is safe to
automate precisely because it needs no judgement: nobody even asserted those.
Step 6b's reviewer prompt says so in as many words, because without that
instruction the reviewer reads the coverage table back to you.

## Design notes

- **Criteria are declared up front**, not accumulated from tags. The useful half
  of the report is the criteria *nothing* claimed, and that is underivable from
  the tags alone.
- **A tag naming an undeclared criterion raises at the call.** Left through, the
  criterion the author meant comes back unclaimed while the storyboard looks
  complete — wrong in the one direction nobody checks. Same posture as
  `caption()` refusing a secret.
- **`ac` is absent from a beat that claims nothing**, like `error` on a beat
  that returned (#24): a take recorded before this key existed reads exactly
  like an untagged beat.
- **A stitched demo recomputes coverage over the merged beats** against the
  union of the segments' criteria. Unioning the segments' own reports would
  point a reviewer at beat numbers absent from the file they are reading, and
  could never name a criterion *no* segment claimed — neither segment knows the
  other's ticket. Conflicting wording for one id is reported, not resolved.
- `coverage` is `null` without `criteria=`. An empty report reads as a take
  measured against a ticket and found to cover none of it.

## Fault injection

`tests/smoke --coverage-only` against five broken copies. Each aborts unless its
pattern matches **exactly once**.

```
--- break: nothing is ever reported unclaimed                      PASS (exit 1)
      | - coverage: `unclaimed` is [], expected ['AC-3']…
      | - coverage-merge: `unclaimed` is [], expected ['AC-3']…
--- break: the report points one beat past the beat that claimed it  PASS (exit 1)
      | - coverage: AC-1 points at beat 5 (`shot`), whose own `ac` is ['AC-2']
        — the report and the beat disagree about what that beat claims
--- break: a tag naming an undeclared criterion is accepted         PASS (exit 1)
      | - coverage-refusals: a tag naming an undeclared criterion was accepted
--- break: a merged demo is judged against only the first segment's ticket  PASS
      | - coverage-merge: the merged criteria are ['AC-1', 'AC-3'], not the
        union of the segments'…
--- break: timeline.md stops stating the finding                    PASS (exit 1)
      | - coverage: timeline.md never states the finding…

All 5 injections were caught.
```

### One assertion was rewritten because the injection showed it graded nothing

The markdown arm originally asserted `"AC-3" in timeline.md`. **AC-3 appears on
its own table row whatever the report concludes**, so it passed on a document
that had dropped the finding entirely. It now looks for the sentence that
*states* the finding and requires the criterion to be named in it — and
likewise checks for `"not what it proved"` rather than for the word `"claimed"`,
which is a column header.

Second time in two PRs that an assertion I wrote was green on healthy and green
on broken, and both times only the injection said so.

## Correspondence, not count

`check_coverage` asserts every beat the report points at really carries that
tag, at that timestamp, naming that still, **in the same file**. "3 criteria
produced 3 rows" is nearly free to satisfy and says nothing — the off-by-one
injection above is what that arm is for.

`check_coverage_refusals` runs without a browser, and every refusal has a
control beside it (a valid tag, a duplicate that must de-duplicate, an untagged
beat) so the refusals cannot pass on a function that rejects everything.

## Stated limits

- **Whether a tagged beat *shows* what it claims is not graded here**, by
  design. That is the reviewer's judgement; the artifact is written to say so.
- **The conformance reviewer is a documented prompt, not code.** Nothing in CI
  runs it. #64 is where it gets run for real against a ticket.
- **No demo video.** A coverage report is read, not watched, so this fails the
  demoable test in `AGENTS.md` — a video of it would be ceremony. The report
  above is the artifact to review.


rogvid:
## Two CI failures, diagnosed

**1. mypy — a real gap, fixed.** `TerminalRecorder` overrides `shot()` to pump
pending output first, so it was dropping the new `ac=` argument entirely: a
terminal take could not tag a still at all. Not a signature nit — a criterion
demonstrated on the command line is the ordinary reason a demo has a terminal
half. Fixed and pushed.

**2. smoke `terminal-opening` — pre-existing, and not this branch. Filed as
#128.**

This branch does not touch that path. The whole diff against main for
`terminal.py` is a `criteria=` keyword passed through to `super().__init__`,
and the `tests/smoke` additions are a new phase the terminal arm never reaches.

Measured rather than asserted:

| where | runs | corner at frame 0 | card covers |
|---|---:|---:|---|
| `main` CI (c47e3ad, 908986c, 95f4e44) | 3 | **26** | 2.65–2.85s |
| this machine | 3 | **26** | 2.75–2.80s |
| this PR's run | 1 | **128** | **0.00s** |

The card is dark (≤ 60), the bare background light (≥ 150). 128 is neither — the
card was part way up when recording started. One failure in four observed CI
runs, none in six local ones; runner-only, and the states are separated rather
than noisy around a bar.

I am re-running CI **because the cause is established and filed**, not to see if
it comes back green. If `terminal-opening` fails again on this branch that is
#128 recurring at a higher rate than measured, and it goes on #128 rather than
being re-run again.

rogvid:
## The second smoke failure was mine, and it was a false positive

Different failure from the first, and this one is a defect in an assertion I
wrote for #119 rather than a flake:

```
- web: this take reports no opening hold (content.opening.held is 0.0), so the
  app had already painted when the recording started and nothing here graded
  the fix for #119.
```

`check_opening` refused to grade a take whose `held` was zero and reported that
refusal as a **failure**. But a run where Chromium's screencast emits its first
frame *after* the page painted is a perfectly healthy recording — it is the
outcome the fix exists to produce. So the check went red on a good take, on a
branch that has nothing to do with the opening.

**The acceptance criterion is about the picture, not the mechanism**, and I
conflated the two. A take whose app painted before frame zero satisfies it by
itself; one whose app did not satisfies it because the hold covered the gap.
Either way frame zero has to show the app, so that arm now runs
unconditionally.

What is genuinely lost is graded coverage of the hold *code* on a run that never
needed it, and the honest handling is to say so rather than to fail. The
summary line now reports which of the two ways the take passed:

```
smoke: web opens on the app (frame 0 scores 19.29 against a 17.63 median,
       over a 0.35s hold)
smoke: web opens on the app (frame 0 scores 19.29 against a 17.63 median,
       with no hold needed — the app had already painted when the screencast
       started, so the hold code was not exercised this run (#126))
```

A run whose evidence is thinner announces it instead of reading identically to
one that proved something. Making the gap reliable needs a fixture that defers
its first paint deliberately — that is #126, and it now closes two things.

### The weakened assertion was re-injected

Removing an arm means re-proving the rest still catch the break. All four
opening injections, against this branch:

```
--- break: the hold is composited but switched off              PASS (exit 1)
      | frame 0 scores 0.00 against a 15.99 median … held = 0.3
--- break: the opening is never detected                        PASS (exit 1)
      | frame 0 scores 0.00 … held = 0.0 — it saw nothing to cover, and the
        app is not there either, so the opening was blank and went unnoticed
--- break: the recorder stops running its opening hold at all   PASS (exit 1)
      | content.opening.held = null … the mechanism is not wired in
--- break: the blank floor is removed                           PASS (exit 1)
      | a static painted picture measured a gap of 1.95, not 0.0

All 4 injections were caught.
```

The second injection is now caught by the **picture** arm rather than by a
premise that also fired on healthy runs, and the third arm is new — `held is
null` means the hold never ran, which is a failure however good frame zero
looks. Net: one arm that could fail for a non-defect removed, two that grade
the criterion added.

Issue #138, opened 2026-07-28
Design: demo-video does not record secrets — delete the masking surface (8,189 lines)


rogvid:
Proposed by the maintainer after a three-way architecture review. This is the
design record; it is **not** a work item. Split it into vertical slices before
anyone starts.

## The measurement that motivates it

**29 of 118 issues — 25% of this tracker — are redaction and secrets.** That is
the largest cluster by subject matter, larger than any single component.

Every one of them was found the same way: a fix for the previous channel
revealed the next. The channels, in the order they were discovered — captions,
stills, frames, the TTS cache, the PTY stream, terminal wrap, alt-screen
switches, line-erase escapes, anchored fragments that never complete a match,
`url`, `title`, the ARIA tree, `outerHTML`, evidence field budgets, the browser
top layer, elements with no stable CSS handle, same-origin iframes, segment
boundaries, the failure dump, and the exception message.

That is an enumeration of every path by which a string can reach a pixel. It
grows whenever a browser or a terminal gains a feature. **#68 is still open and
is the worst shape available: a mask that verifies and leaks into `demo.mp4`
anyway.**

## The argument

A masking system that works 99% of the time is worse than none, because it is
what convinces someone it is safe to point the recorder at production. The
absence of the mechanism is a stronger guarantee than the mechanism.

Real credentials should not be in a CI pipeline in the first place, and this
repo's own workflow already refuses to record against a public host. Dummy
values in a fixture are not secrets — they are examples, and nothing needs to
hide them.

So: **if a demo needs a real secret on screen, the answer is that it cannot be
demoed** — not that the recorder needs one more mask.

## The change

Split two things that currently share a codepath. Only one of them goes.

### `register_secret(value)` — becomes a tripwire, not a mask

- Stops masking. **Delete the masking machinery**: the scrub pipeline, the
  elastic-regex evidence family, the terminal scrubber's escape-class handling,
  the per-artifact masking passes.
- Becomes one sweep across every artifact for the literal, after they are
  written and before they are released.
- On a hit: **the take fails and deletes its output.** No mp4, no frames, no
  evidence, no timeline.

Why this is the right shape: "does this string appear in these bytes" is
**bounded and checkable in a single pass**, and it is an assertion that can be
*seen to fail*. "Have I covered every rendering path" is unbounded and cannot.

### `redact(selector)` — keep it, unchanged in spirit

This is a different risk class and a legitimate one: a staging database full of
[...]

Issue #136, opened 2026-07-28
tests/smoke has no injection manifest, so the discipline is executed for the 720-line suite and prose for the 12,369-line one


rogvid:
The largest category in this tracker is *measurements that do not grade what
they claim*. This issue is why that category exists, and it is a cost problem,
not a care problem.

## The asymmetry, verified

`tests/ci-unit` **executes** the fault-injection rule:

- `tests/ci-unit:530` — a 12-entry `INJECTIONS` table, each entry
  `(name, script, old, new, must_fail)`.
- `tests/ci-unit:632-657` — a driver that applies each break, requires the
  suite to go red, and requires the **named** assertions to be the ones that
  fired: `noticed = {t for t in must_fail if t in done.stderr}`.
- `tests/ci-unit:679` — `f"{failures} injection(s) went unnoticed — those
  assertions grade nothing."`
- `.github/workflows/ci.yml:87-88` — it runs on every push.

`tests/smoke` contains the string `INJECTION` **zero times**. Its injections
exist as past-tense markdown tables in `tests/README.md:1488` and `:1535`.

So the 720-line suite has an executed manifest, and the 12,369-line suite that
verifies the 8,849-line recorder has a changelog.

## Why, measured on one box

| | |
|---|---|
| one `ci-unit` injection | ~0.3 s |
| `tests/smoke --coverage-only` | 12 s |
| `tests/smoke --content-only` | 128 s |
| `tests/smoke` (full) | **10 m 22 s** |
| concurrency | refused — `only_one_suite`, `tests/smoke:12126` |

`only_one_suite` exists because of #78: two timing bars fail under contention.
So injections cannot be parallelised on one machine either. At 12-620 s each
under an exclusive lock, nobody maintains a manifest. They write a table, once,
and the table stops being true.

That drift is what the issue titles say out loud: #82 *"the prose and the run
line say all of them"*, #107 *"plants 2 of the 4 filenames"*, #87 *"that sweep
now goes through `leaked_in()`"* against a line byte-identical to `main`.

## What a re-run found

Six injections against `--coverage-only`, 72 s total: **five caught**. That arm
is sound, and it is the arm built on constructed inputs
(`check_coverage_merge`, `check_coverage_refusals`). Reporting that plainly
because it is the result.

One survivor, filed separately. And one injection against `--content-only`
found a blocking-class hole — #135.

The two real findings came from injections **nobody had written down**. That is
the argument for a list that grows monotonically and stays executed: a markdown
[...]

Issue #145, opened 2026-07-28
Retire the secret fixture, the gitleaks allowlist and the gitleaks CI job


rogvid:
Slice 5 of #138, and the last. Small, and it goes after #144 so nothing is
removed while something still reads it.

## Delete

- **`tests/fixture/index.html`** — the secret panel and the `?secret=1` /
  `?secret=closed` hooks. Roughly **266 of 904 lines**, measured by keyword
  cluster, so ±20. Verify against the file rather than trusting the number.
- **`.gitleaks.toml`** — 155 lines, largely the fixture allowlist. Its narrow
  allowlist existed to let the fixture's fake `sk-live-` values through; with the
  fixture panel gone there is nothing to allow.
- **The `secrets (gitleaks)` job** in `.github/workflows/ci.yml`, ~22 lines.

## The judgement to make explicitly, not by default

Dropping the gitleaks job is the one item here that is not purely mechanical.
The argument for dropping it: it exists in this repo because the fixture
deliberately contains secret-shaped strings, and after #144 it does not.

The argument against: it is a repo-wide guard against a real credential being
committed by accident, which is unrelated to demo-video's fixtures and is cheap
(10 s). **My read is that it should stay** on those grounds and only the
fixture-specific allowlist in `.gitleaks.toml` should shrink to nothing — but
this is the maintainer's call, and it should be made rather than inherited.

Note the asymmetry that makes it worth a thought: #138 removes demo-video's
ability to protect a secret *in a recording*. It says nothing about protecting
one *in the repository*, and conflating the two would be exactly the confusion
the boundary statement in #141 exists to prevent.

## Acceptance

`tests/smoke` passes with the fixture's secret panel gone, no surviving arm
requests `?secret=`, and whichever decision is taken on gitleaks is recorded in
the commit message with its reason.

## Deliberately not included

- `demo-target-guard`, the egress guard that refuses to record against a public
  host. It is unrelated to masking — it stops the recorder being pointed at
  production, which under #138's policy matters *more*, not less. It stays.

Issue #144, opened 2026-07-28
Delete the core secret primitives and the evidence masking family


rogvid:
Slice 4 of #138. **938 of `core.py`'s 5,344 lines**, plus the smoke arms that
grade them. `core.py` ends at roughly **4,406 lines**.

**Depends on #142 and #143 landing first** — this removes the primitives those
slices' code calls, so it goes last among the deletions.

## Delete

| what | where | lines |
|---|---|---:|
| `# -- secrets`: `SECRET_MASK`, `SECRET_MIN_LEN`, `SecretLeak`, `Secret` | `367-454` | 88 |
| `EVIDENCE_HTML_WITHHELD`, `EVIDENCE_WRAP_MIN_TOKEN`, `_JSON_*` escapes, `EVIDENCE_MIN_MASK_LEN` | `893-935` | 43 |
| Evidence-header and failure-section leak commentary | `851-865`, `1431-1451` | 36 |
| `__init__`: `_secrets`, `_redacted`, `_mask_seen`, `_nav_pending`, `_last_sync`, `_redact_style`, `_shots`, `_evidence_masks`, `_evidence_outside`, `_evidence_res` | `2737-2800` | 36 |
| Base `_checkpoint` + `_before_shot` | `2882-2894` | 13 |
| `register_secret` / `secrets` / `scrub` / `_verify_redaction_final` / `_scrub_deep` / `_no_secrets` | `2895-3018` | 124 |
| `__exit__` verify block, `unmasked`, `leaked`, discard branch, marker branch, raise branch | `3093-3360`, 11 sub-ranges | 136 |
| `_beat` scrub commentary and `scrub()` calls | `3596-3679` | 22 |
| `_capture_evidence` SecretLeak handling | `3698-3711` | 10 |
| `_evidence_forbidden` … `_evidence_scrub_deep` (7 methods incl. `_evidence_re`, `_evidence_probe`, `_recoverable_secret`, `_evidence_holds`) | `3727-3909` | **183** |
| `_evidence_doc` masking, withheld-markup, byte backstop | `3910-3989` | 52 |
| `_failure_doc` / `_build_failure` / `_write_failure_marker` / `_clear_failure_dir` masking passes | `4124-4453` | 30 |
| `_timeline_doc` scrub commentary and all four scrub passes | `4595-4680` | 48 |
| `_discard_artifacts` (whole function; only caller is the leak branch) | `4825-4879` | 55 |
[...]

PR #146, opened 2026-07-28
Give the recorder a check faster than ten minutes, and SKILL.md a context budget

### LINKED ISSUE #14: demo_recording is Unix-only even for web-only demos

### LINKED ISSUE #139: core.py's browser-free half is unreachable without Chromium, so the recorder has no check faster than 10m22s
## The measurement

`core.py` is 5,344 lines. **2,580 of them contain not one reference to
Playwright** — verified by grep over the whole file:

```
$ grep -n "playwright\|sync_playwright\|Page" core.py
43:from playwright.sync_api import Page, sync_playwright
2718:        self.page: Page = None
3024:        self._pw = sync_playwright().start()
3383:  (a docstring)
3459:    def _watch_page(self, page: Page) -> None:
```

Every hit outside line 43 is inside `_DemoBase` (lines 2592-4999). The rest —
`render_timeline_md`, `write_timeline`, `coverage_report`, `_merged_coverage`,
`content_report`, `merge_content`, `opening_gap`, `opening_report`,
`render_frames_md`, `scene_times`, `beat_frames`, `_check_stream_shapes`,
`_merged_timeline`, `stitch` — are pure functions of dicts and paths.

They are unimportable anyway:

```
$ python3 -c "import demo_recording.core"
ModuleNotFoundError: No module named 'playwright'
```

## Why it matters, and it is not tidiness

This is the cost that makes #136 true. The fault-injection rule is *executed*
for `tests/ci-unit` (720 lines, ~0.3 s per injection) and *prose* for
`tests/smoke` (12,369 lines, 10 m 22 s, exclusive `flock`). #136 measures why:
at 12-620 s per injection under a lock nobody maintains a manifest.

**The recorder does not have a harness the discipline is affordable in.** Half
of it does not need one built — it needs the import moved.

## Acceptance criterion

`from demo_recording.timeline import render_timeline_md` (and the other
browser-free modules) succeeds in an interpreter with no third-party package
installed, and `tests/unit` grades those modules in under a second with an
executed injection manifest that refuses to pass when a break goes unnoticed —
the `tests/ci-unit` driver's contract, not a markdown table.

## Deliberately out of scope

- `_DemoBase` itself. Splitting its seven concerns is a separate slice and
  cannot be attempted safely until this lands.
- Skipping any smoke take. The browser suite keeps every job only it can do.
- #61's fast/full CI split. This makes it possible; it does not do it.

## Related

- #136 — the injection manifest asymmetry this is the precondition for fixing
- #61 — smoke is approaching 4 minutes (now 10m22s); fast job vs full job
- #14 — `demo_recording` is Unix-only even for web-only demos, same import graph

### LINKED ISSUE #140: SKILL.md costs 40,510 tokens to answer "record a demo of X", and ~74% of it is not needed to write a storyboard
## The measurement

`skills/demo-video/SKILL.md` is **1,893 lines ≈ 40,510 tokens**. A skill's
SKILL.md is loaded *whole* into the caller's context the moment the skill
triggers; referenced files are read on demand. So recording a 30-second web
demo with no secrets, no segments, no terminal and no ticket costs a fifth of a
200k window before the first verb is written.

Lines by section:

| section | lines |
|---|---|
| Redacting secrets | 320 |
| The beat timeline (+ `content`, `opening`) | 294 |
| Per-beat evidence | 192 |
| Process | 163 |
| Terminal demos | 122 |
| Determinism | 119 |
| Failing the take on a broken app | 80 |
| Recording against a ticket | 78 |
| Recording on a pull request (CI) | 74 |
| When a take does not finish | 72 |
| Review frames | 72 |
| everything else (overview, setup, template, config, API, pacing, mistakes, sharing) | 307 |

**~490 lines are needed to write any storyboard. ~1,400 are needed only once
the author hits the thing they describe.**

## The shape, in the repo's own terms

The content is not the problem — it is the best-argued document here and every
limit in it was earned. The problem is that the **interface** (what an agent
must know to write a storyboard) is buried inside the **implementation** (why
every control is shaped the way it is). That is the same shallowness as a wide
`__all__`, in prose.

## Acceptance criterion

`SKILL.md` is under ~500 lines, and `examples/ticket-queue`'s storyboards can be
re-derived from the trimmed file alone — every section that moved leaves an
explicit read-pointer at the point of use (`Read reference/secrets.md before the
first redact()`), because an unread reference is a deleted one. Nothing is
deleted; the argued limits move one hop away.

## The line that decides each section, and it is not length

A section stays if an agent that never reads it would write a **wrong**
storyboard. It moves if the agent only needs it once it hits the thing.
"Register before you caption" is the first kind. The three-second
shape-fragment ceiling is the second.

## Deliberately out of scope

- Rewriting any of the prose. This is motion plus pointers.
- The frontmatter `description`, which is what an agent matches on to trigger
  the skill at all. Unchanged.
- #138's rewrite of the secrets surface. If that lands, it rewrites one
  reference file instead of a third of SKILL.md — which is an argument for
  doing this first, not a conflict.

rogvid:
Two slices that share no file, taken together because they are the two places
this skill is paying a cost on every use: it cannot be verified quickly, and it
cannot be loaded cheaply.

Closes #139. Closes #140. Closes #14.

---

## Slice A — cut core.py at the line the browser already draws (#139)

**Acceptance criterion:** the browser-free modules import in an interpreter with
no third-party package installed, and `tests/unit` grades them in under a second
with an *executed* injection manifest that refuses to pass when a break goes
unnoticed.

### The measurement that motivates it

`core.py` was 5,344 lines, and **2,580 of them contained no reference to
Playwright at all**. Every hit outside line 43 was inside `_DemoBase`:

```
$ grep -n "playwright\|sync_playwright\|Page" core.py
43:from playwright.sync_api import Page, sync_playwright
2718:        self.page: Page = None
3024:        self._pw = sync_playwright().start()
3383:  (a docstring)
3459:    def _watch_page(self, page: Page) -> None:
```

The consequence is the finding, not the import: the recorder had no check
faster than `tests/smoke` — 12,369 lines, ten minutes, exclusive `flock`. That
is the cost #136 measures when it finds the fault-injection rule *executed* for
the 720-line `ci-unit` suite and written down as prose for the 12,369-line one.

### What moved

| module | lines | what it owns |
|---|---:|---|
| `secrets.py` | 98 | `Secret`, `SecretLeak`, and the constants that bound them |
| `coverage.py` | 179 | acceptance criteria, and what a storyboard claimed |
| `timeline.py` | 668 | the beat log's schema and its two renderings |
| `failure.py` | 75 | where a take that did not finish writes itself down |
| `frames.py` | 363 | beat-aligned review frames |
| `content.py` | 842 | did the recording show anything |
| `stitching.py` | 356 | joining segments losslessly |
| `core.py` | 2,921 | `_DemoBase`: the browser, narration, ffmpeg |

**The motion is pure.** All 108 top-level definitions survive with
byte-identical ASTs, verified by parsing the original and the eight files and
checking for lost, gained, altered and duplicated names:

```
definitions before: 108   after: 108
lost:     none
gained:   none
CHANGED:  none — every moved body is byte-identical in AST
duplicated across modules: none
```

The dependency graph is acyclic by construction. `_common` moved to `content.py`
because leaving it with `stitching.py` was the one cycle in the first pass.

### `__init__` resolves the recorders lazily, which closes #14

`from demo_recording import Recorder` — the *web* recorder, which has no PTY
dependency — used to fail at import on Windows because `terminal.py` imports
`fcntl` at module scope. `SKILL.md` has always scoped the Unix-only constraint
to terminal recording, so the code was stricter than the documentation promised.
PEP 562 `__getattr__` fixes that and is also what makes the browser-free surface
importable with nothing installed.

```
$ python3 -c "import demo_recording; ..."
package imported. playwright loaded? False
all browser-free modules imported. playwright loaded? False
```

### `tests/unit`

52 assertions, **0.07 s**, `dependencies = []`. It does not reduce what smoke
has to run; it runs the half that never needed a browser. What it deliberately
does **not** cover is stated at the bottom of the file.

---

## Slice E — SKILL.md's reference material moves behind read-pointers (#140)

**Acceptance criterion:** `SKILL.md` under ~500 lines, nothing deleted, every
moved section linked at the point of use.

A skill's `SKILL.md` is loaded **whole** into the caller's context the moment
the skill triggers; linked files are read on demand. This one was 1,893 lines —
**40,510 tokens, a fifth of a 200k window** — spent before the first storyboard
verb is written.

| | lines | chars | tokens |
|---|---:|---:|---:|
| `SKILL.md` before | 1,892 | 111,941 | ~40,510 |
| `SKILL.md` after | 510 | 32,215 | **~11,658** |
| `reference/` (8 files) | 1,416 | 85,745 | ~31,030 (on demand) |

**Per-invocation cost falls 71%. Total prose is 105% of the original** — the
extra 5% is the header each reference file carries. Nothing was deleted.

The line that decided each section is not length: a section stays if an agent
that never reads it would write a *wrong* storyboard, and moves if the agent
only needs it once it hits the thing. "Register before you caption" is the first
kind; the three-second shape-fragment ceiling is the second.

Four cross-references the move would have broken are repointed at the files
those sections now live in.

---

## Fault injection — the failure output

Per the house rule. Each entry breaks one named thing, requires the **named**
tests to go red, and aborts rather than passing when its pattern does not match
exactly once.

```
PASS  break: coverage accumulates criteria from the tags instead of the declaration
PASS  break: a take with no criteria reports an empty coverage object instead of null
PASS  break: a claim records only the beat index, losing the frame it points at
PASS  break: segments disagreeing about a criterion's wording are silently resolved
PASS  break: the criterion id pattern stops excluding a pipe
PASS  break: the table cell stops escaping pipes
PASS  break: a beat with no timestamp renders as time zero
PASS  break: the coverage section calls a claim a demonstration
PASS  break: the unclaimed sentence stops naming which criteria are unclaimed
PASS  break: a segment's timeline is written to the whole demo's path
PASS  break: a recorder that never held its opening warns like one that did
PASS  break: the opening warning fires below its own bar
PASS  break: the held-stretch correlation counts a beat that began before the stretch
PASS  break: an acting verb is classified as narration
PASS  break: the content rect keeps the caption band it exists to exclude
PASS  break: a merged demo reports the best segment's score instead of the worst
PASS  break: a merged demo reports the shortest held stretch instead of the longest
PASS  break: a merged warning loses the segment it came from
PASS  break: a reference link in SKILL.md points at a file that is not there
PASS  break: a whole section drifts back into the always-loaded SKILL.md
PASS  break: a reference file stops being linked from SKILL.md
PASS  break: Secret becomes a str subclass, so a verb can log one as its target
PASS  break: a Secret's repr discloses the value
All 23 injections were caught.
```

The abort guard was itself verified, by planting a pattern that is not there:

```
ABORT: injection 'coverage accumulates criteria from the tags instead of the
declaration' matched 0 times in coverage.py, expected exactly 1. The injection
did not land, so nothing it would have proven is proven.
$ echo $?
3
```

### Two defects the discipline found, in this change

1. **A blanket mypy suppression was hiding a second diagnostic.**
   `[mypy-demo_recording.core] disable_error_code = arg-type` exists for one
   *stated* reason (a `ViewportSize` TypedDict). Moving `_fmt_t` out from under
   it surfaced an unrelated `float(object)` error it had been silencing for
   free. Fixed with a `cast` — erased at run time, so behaviour is unchanged —
   rather than by spreading the suppression to seven new modules. **This is the
   one edit in slice A that is not pure motion.**

2. **My own injection was mis-targeted.** The break that removes a reference
   link first targeted `ci.md`, which `SKILL.md` links *twice* — so the
   assertion stayed silent against broken input and the manifest reported
   `1 injection(s) went unnoticed`. Retargeted at `narration.md`, the one file
   linked exactly once.

---

## Verification

| check | result |
|---|---|
| `ruff check .` / `ruff format --check .` | pass (11 files) |
| `mypy skills/demo-video/helpers/` | pass, 11 source files |
| `actionlint .github/workflows/*.yml` | pass |
| `tests/ci-unit` + `--fault-inject` | pass, 12 injections caught |
| `tests/unit` | **52 assertions, 0.07 s** |
| `tests/unit --fault-inject` | **23/23 injections caught** |
| `tests/smoke` (full) | **PASSED — 133 assertions, exit 0** |
| three committed `ticket-queue` storyboards | imports resolve |

`tests/unit` is wired into CI as its own job, deliberately **without**
`--with playwright`: if that job ever needs it, the import boundary has been
broken again and the step is what says so.

---

## Stated limits, and what this deliberately leaves alone

- **`_DemoBase` is untouched.** It is still 2,408 lines, 73 methods and 55
  instance attributes, and its seam still leaks: `Recorder` reaches into 15 of
  the base's private attributes and `TerminalRecorder` calls 10 of its private
  methods, beyond the 11 hooks the base declares. Splitting it is the next
  slice and could not be attempted safely until this one existed. Filed
  separately rather than smuggled in here.
- **`__all__` is unchanged at 50 names**, 23 of which have no caller anywhere
  in the repository. Narrowing it is a separate slice; this PR changes which
  module each name comes from, not which names exist.
- **`web.py` is still 64% JavaScript in string literals** (781 of 1,226 code
  lines, including all 323 lines of `_REDACT_JS`), which no linter or syntax
  check can read. Left alone deliberately: #138 proposes deleting much of that
  surface, and churning it first would be work on code slated for removal.
- **`tests/unit` does not verify that the split preserved behaviour.** The
  byte-identical AST comparison and the full `tests/smoke` run are what do
  that. A green `tests/unit` says the pieces are right, not that they are still
  wired together.
- **Nothing here measures real frames.** `content_report`'s ffmpeg sampling,
  `opening_gap`, `_luma_stddev` and `_moved_pixels` need a video to be true
  about; what is graded is the logic around them.
- **`SKILL.md`'s prose is unrewritten.** This is motion plus pointers. Where a
  moved section's internal wording says "above" or "below" about material that
  is still in the same file, it is still correct; the four places where it was
  not are fixed.

## Not demoable

Per the house test: none of this can be verified by watching. It is an import
boundary, a test harness and a documentation split — verifiable only by reading
a file, an assertion or a log. No demo video.


rogvid:
**Sequencing note — this lands ahead of #138's slices, deliberately, and against the recorded order.**

#138's comment (08:35) says #138's slices go first and not concurrently with
#139. I had read #138 at ~08:2x, before that comment existed, and opened this
PR at 08:59 without it. Recording the deviation rather than leaving it implied.

The maintainer's argument and what changed:

- **"Both are large edits to the same three files."** That was a concurrency
  argument, and it dissolves on merge rather than on hold: #142–#144 then delete
  from seven small modules with a 0.07 s suite under them, instead of from a
  5,428-line `core.py` whose only check takes ten minutes.
- **"#140 first means relocating 474 lines into reference files these slices
  then delete."** True, and that cost is now sunk. It also inverts: the whole
  320-line `## Redacting secrets` section is now `reference/secrets.md`, a
  `git rm` rather than a 320-line excision from a 1,893-line file.
- **"Extracting `Secret`/`SecretLeak` and then deleting them is wasted
  motion."** Correct, and unrecovered: `secrets.py` is 88 lines that #144
  deletes. Accepted as the cost of this ordering.

Slices #141 → #145 follow on top of this, in the order #138 specifies.

Issue #143, opened 2026-07-28
Delete the terminal PTY scrubber


rogvid:
Slice 3 of #138. **572 of `terminal.py`'s 1,389 lines (41%)**, plus the 839-line
smoke block that grades them. `terminal.py` ends at roughly **817 lines**.

Depends on #141. Independent of #142 (web) — they touch different files — but
both must land before #144 removes the core primitives they call.

## Delete

| what | where | lines |
|---|---|---:|
| PTY-scrubber block: 98-line design comment, `_INERT_ESC_RE`, `_PARTIAL_ESC_RE`, `SECRET_SHAPES`, `_ANCHORED_HOLD_S`, `_WITHHOLD_WARN_S`, `_MAX_HOLD`, `_MIN_DANGLING_PREFIX`, `_strip_inert`, `_merged_spans`, `_StreamRedactor` (198) | `terminal.py:251-675` | **425** |
| `_verify_redaction_final` | `terminal.py:1042-1103` | 62 |
| `_warn_if_withholding` | `terminal.py:1016-1041` | 26 |
| `_pump` redactor wiring, `_write_out` docstring, `_stop` flush, `_redactor` / `_withheld_warned` init | scattered | 23 |
| `send(Secret)`, `run`, `key`, `_open_on_card` `_no_secrets` calls, docstrings, imports | scattered | 36 |

`send()`'s signature narrows from `str | Secret` to `str`.

## Also delete, in the same change

- The terminal-redaction smoke block (`tests/smoke:9403-10241`, **839 lines**)
  and the terminal-redaction constants (part of `1463-1930`).
- `run_terminal_wrap` (`tests/smoke:8236-8377`, 142). Note this closes **#86**
  by deletion — its complaint is that the take's geometry premise does not
  observe the wrap it claims to test, and the take exists to prove a wrapped
  secret is masked.
- `SKILL.md` `### Redaction in a terminal demo` (inside the `330-649` block, so
  it goes with #142 if that lands first — check before duplicating) and
  `tests/README.md` Terminal redaction write-up (`928-1104`, 177).

## Carve-out

**`_StoryboardFailed` (`tests/smoke:10242-10248`) must be relocated, not
deleted.** It sits inside the terminal-redaction block but its callers are
`run_crash_between` (`11571`, `11579`), which survives. Move it before deleting
the block around it.

## Acceptance

`tests/smoke` passes with no terminal-redaction arm present, `terminal.py` is
[...]

Issue #142, opened 2026-07-28
Delete the web masking surface: redact(), the paint gate, and the evidence harvest


rogvid:
Slice 2 of #138. The largest single slice: **~1,238 of `web.py`'s 2,000 lines**,
plus the smoke arms that grade them and the SKILL.md section that documents
them. `web.py` ends at roughly **760 lines**.

Depends on #141 (the boundary statement) landing first. Must land before slice 4
(#144), which removes the core primitives these call.

## Delete

| what | where | lines |
|---|---|---:|
| Redaction design comment, `REDACT_*` constants, `_REDACT_GATE_ID` | `web.py:276-351` | 76 |
| `_REDACT_JS` | `web.py:353-676` | 324 (JS) |
| `_CSS_ONLY_JS` | `web.py:678-685` | 8 |
| `_EVIDENCE_HARVEST_JS` + header — feeds the mask only | `web.py:687-939` | 253 |
| `_EVIDENCE_HTML_JS`: `hide`, `isRedacted`, ancestor walk, elide loop | `web.py:940-1034` | ~50 |
| `EVIDENCE_HARVEST_TRIES`, `EVIDENCE_OUTSIDE_MAX` | `web.py:1042-1052` | 11 |
| `redact` … `_verify_redaction_final` (8 methods) | `web.py:1273-1585` | 313 |
| `_failure_screen` harvest half → `return self._failure_page_text()` | `web.py:1587-1663` | ~57 |
| `_before_shot`, `_note_navigation`, `_harvest` | `web.py:1675-1715` | 38 |
| `_capture_page` html opts, `_evidence_payload` → `_capture_page()` | `web.py:1745-1816` | 58 |
| `goto` paint-gate block, six `_checkpoint()` calls, `type_into` Secret branch, imports | scattered | 36 |

**`redact()` goes outright** — decided by the maintainer over keeping it as a
visual region blur. Keeping the pixel half would have retained ~342 lines and
saved ~896 instead of 1,238. The reason for deleting: under the stated policy a
staging database of real customer data is not something this tool records
either, so the remaining use case is aesthetic, and it removes the last thing
that could be mistaken for a security control.

**`_checkpoint` is masking-only** — verified, not assumed. It returns
immediately when `self._redacted` is empty and the base is a no-op. Its six verb
call sites (`goto`, `spotlight`, `move_to`, `click_fast`, `scroll_to`,
`wait_for`, `_idle`) delete cleanly.

[...]

Issue #141, opened 2026-07-28
Say what demo-video records, and what it does not defend against


rogvid:
Slice 1 of #138, and it ships **before** any deletion.

## Why first

#138 deletes 8,189 lines of masking, scrubbing and redaction. If the code goes
before the sentence, the absence reads as an oversight and the next person
helpfully reinvents it — which is how the surface got here.

This is the #63 move applied to something much larger, with the warning from
#123/#124 attached: a boundary statement is itself an artifact, and artifacts
here lie unless something grades them. So the sentence must be one a reader can
check against behaviour, not a reassurance.

## What to write

A short section near the top of `SKILL.md`, before the API:

- demo-video records **fixtures and example data**. Videos are generated in CI,
  which is secret-free.
- Values written in a storyboard or a timeline are **examples**. A string that
  would be a credential in production is not one in a fixture, and nothing
  needs to hide it.
- The tool has **no masking, no scrubbing and no redaction**, and does not
  defend against a secret reaching the screen. If a real credential can appear
  in the app you are recording, **do not record it** — that is the whole of the
  guidance.
- Say the corollary plainly: a demo that requires a real secret to be
  meaningful is a demo that cannot be made. That is a constraint on what gets
  demoed, not a gap to be filled later.

Delete the two verb-table rows for `redact()` and `register_secret()` in the
same change if slice 2 has not landed yet; otherwise they go with their code.

## Acceptance

A reader who reaches the API section knows, without reading further, that this
tool records non-production data and offers no protection. The claim is stated
as a scope limit, not as a security property, and nowhere in `SKILL.md` does a
surviving sentence promise that anything is hidden — checked by grepping the
file for the claim words (`redact`, `mask`, `scrub`, `secret`) and confirming
every remaining hit is either this section or a deletion still pending.

## Deliberately not included

- Any code change. This slice is prose only, and it is the only slice that is.
- The rest of #140's restructuring of `SKILL.md`. This section is written where
  it will still be correct after #140 moves things around it.

rogvid:
Shipped in #149. The section sits above the Overview in SKILL.md, and #142/#143/#144 have since made it true of the code as well.

Issue #61, opened 2026-07-25
tests/smoke runs whole on every push at ~8 minutes, and its arms are now individually addressable


rogvid:
**Re-scoped 2026-07-31.** Every number in the original was stale and the fast
tier it asked for already exists.

| the original said | today |
|---|---|
| "approaching 4 minutes" (3m41s) | **~7m30s** on CI |
| "split it into a fast job and a full job" | a browser-free `unit` job exists (#146) — 82 tests, ~15 s |
| "don't skip the secrets/redaction takes" | those takes were deleted (#138) |
| — | `ci.yml`'s step name advertised redaction takes until #166 |

## What actually remains

`ci.yml` still runs the **whole** `tests/smoke` on every push, with no arm
selection, and the suite has kept growing — it is 22 takes now, and #170 added
another.

The situation changed in a way the original could not anticipate: **the arms are
individually addressable and individually costed.** Measured while building
#174's manifest, on one machine, one arm at a time:

```
evidence   7s     determinism 19s    failure  48s
coverage   8s     polish      26s    web     123s
narration  8s     segments    29s    content 148s
strict    13s     overlay     31s    terminal 186s
                                     whole   622s
```

Three arms are 74% of the cost. That is a real split point rather than the
guess the original was making.

## The judgement to make

Not "fast job / full job" — that framing is spent, because the fast job exists.
The question is whether a **push** should record every take, or whether the
cheap arms (evidence, coverage, narration, strict, determinism — 55 s together)
belong on every push with `web`/`content`/`terminal` on merge to `main` and on
a label.

Against splitting: a take that only runs on merge is a take that fails after the
review, and this suite exists to catch the recorder producing an artifact that
lies. #174's manifest is already scheduled that way and is a precedent either
way you read it.

**No acceptance criterion yet — this needs the decision made before the work.**

rogvid:
Blocked on #173, not on a decision.

The case against splitting is that the three expensive arms (\`web\` 123 s, \`content\` 148 s, \`terminal\` 186 s — 74% of the suite) are the ones that catch the recorder writing an artifact that lies, so moving them past review is exactly the wrong trade. That argument only holds while those arms are the *only* thing grading those checks.

#173 is the same three arms seen from the other side: its five ungraded check functions are ungraded *because* their cheapest arm costs 123–186 s. If #173 pulls them into `tests/unit` at sub-second cost, the expensive arms stop being the sole evidence, and the split this issue asks for stops costing anything.

[...]

PR #151, opened 2026-07-28
Move tests/smoke off masking, and write run_evidence its own assertions (#150)

### LINKED ISSUE #150: Move tests/smoke off masking, and rewrite run_evidence to grade evidence without a mask
The re-cut #138 needs, and the new first slice of the deletion. **#142, #143
and #144 all depend on this one and none of them can land before it** — see
the measurement on [#138](https://github.com/rogvid/skills/issues/138#issuecomment-5103787096).

## Why this exists

#138 split the deletion by source file: #142 `web.py`, #143 `terminal.py`,
#144 `core.py`. Those three files really are independent. `tests/smoke`, which
grades all of them, is not:

| what | why it crosses a slice line |
|---|---|
| `native_frames`, `edge_energy` | #142 deletes them as "every caller is redaction". True — and `_term_deltas`, a #143 caller, is one of them |
| `REDACT_SECRET`, `REDACT_CONTROL`, `CLOSED_SECRET` | declared in #142's constant block, consumed by #143's `check_terminal_redaction` and `check_cursor_screen_leak` |
| `run_crash_no_evidence` | a #144-era take that calls `rec.redact()`, which #142 removes |
| `check_refused_take_marker` | asserts a take refused for a secret leak; after #142 nothing can refuse one |
| `check_secret_floor`, `check_evidence_leak_is_fatal` | #144 arms, unreachable once #142 lands |

Deleting #142's source alone leaves the suite with undefined names. Deleting
#143's arms alongside it leaves the terminal scrubber shipped and ungraded for
a commit. Either way a slice is red on its own, which is the one thing the
split existed to prevent.

**So the suite moves once, ahead of the source.** After this lands, the three
source slices are each small, independent, and graded by a suite that no
longer mentions masking.

## Delete

Verified against the tree at the time of writing; line numbers will drift.

- **Web arms** — `check_redaction` (592), `record_redaction` (429), the
  `REDACT_*` constant block (~165), `check_unmatched_redaction` +
  `check_storyboard_leak_discards` + `run_redaction` (185), the tamper /
  gate-attack JS + `stub_narration` + `_blur_px` (~110), `blank_run_s` +
  `_edges` + `MAX_BLANK_RUN_S` (47), `native_frames` + `edge_energy` (87).
- **Terminal arms** — `record_terminal_redaction` (324),
  `check_terminal_redaction` (187), `_term_deltas`, `_check_term_pixels`,
  the cursor-leak block (`TERM_CURSOR_SCRIPT` through
  `check_cursor_screen_leak`, ~215), `run_terminal_redaction`, the `TERM_*`
  secret literals and bars (~224).
- **Core arms** — `check_secret_floor` (140),
  `check_evidence_leak_is_fatal` (95), `run_crash_no_evidence` (113),
  `check_refused_take_marker` (55), `leaked_in`, `find_in_frames`,
  `tts_cache_name`, `TERMINAL_WRAP_SECRET`, `run_terminal_wrap` (142).
- **Runner wiring** — the `--redaction-only` flag, and the `run_redaction` /
  `run_terminal_redaction` / `run_terminal_wrap` / `run_crash_no_evidence`
  phases in `run_phases` and `run_failure`.
- **`tests/README.md`** — the Redaction write-up (~241 lines) and the take
  list's redaction entries.

Measured on a working attempt: **`tests/smoke` 12,369 → ~8,800 lines.**

## Rewrite, do not delete: `run_evidence` (467 lines)

**This is the only part of the slice that needs judgement, and it is why the
slice is worth its own review.** Per-beat evidence is not a masking feature —
it survives #138 intact. What this arm currently grades is mostly the mask:

- the `EVIDENCE_SECRETS` byte sweep over every file the take wrote;
- `EVIDENCE_KEEP` — "this string renders outside everything `redact()` covers,
  so it must still be in the evidence", which is the over-masking control;
- the rotating-value path: a card rewritten every 5 ms must come back
  `{"omitted": …}` because the harvest and the snapshot disagree. **With no
  mask there is nothing to reconcile and `omitted` stops existing** — the
  three round trips collapse to one;
- the straddle / split-value / entity / attribute arms, all about a mask
  finding a value the markup wrote differently.

What should survive, and needs assertions written for it rather than inherited:

- every beat writes an evidence file, named as the beat's `evidence` pointer
  says, and the timeline's pointers resolve;
- the ARIA snapshot is present and non-trivial — the catalogue's vacuous-sweep
  guard applies, since an empty snapshot would satisfy most of what is left;
- `scope` / `scope_aria` / `html` are the spotlight target's and null when no
  spotlight is up;
- `_EVIDENCE_HTML_JS` still strips `<script>`, `srcdoc` and value-bearing
  attributes — **these survive #142 and are the one part of that JS with a
  reason unrelated to secrets**, so they need a check that did not previously
  exist on its own;
- truncation is marked, never silent, at the stated budgets;
- a segment's evidence is named `<segment>.seg.beat-NN.json`.

## Acceptance

`tests/smoke` passes with no masking arm present and the string `redact`
appearing nowhere in it; `--redaction-only` is not a flag; `run_evidence`
grades per-beat evidence on its own terms, including a non-triviality control
on the ARIA snapshot and a check that the markup serializer still strips
script text and value-bearing attributes.

## Deliberately not included

- Any change to `web.py`, `terminal.py` or `core.py`. Those are #142, #143 and
  #144, and each becomes small once this lands.
- `.gitleaks.toml`, the gitleaks CI job and the fixture's secret section (#145).
- `tests/unit`. It never referenced masking.

rogvid:
Slice 1 of the re-cut #138, and the precondition for #142, #143 and #144.

Closes #150.

## Why the suite moves first

#138 split the deletion by source file. Those three files are independent;
`tests/smoke`, which grades all of them, is not:

| what | why it crosses a slice line |
|---|---|
| `native_frames`, `edge_energy` | #142 deletes them as "every caller is redaction" — true, and `_term_deltas` is one of those callers, in #143 |
| `REDACT_SECRET`, `REDACT_CONTROL`, `CLOSED_SECRET` | declared in #142's constant block, consumed by #143's `check_terminal_redaction` and the cursor-leak arms |
| `run_crash_no_evidence` | a #144-era take that calls `rec.redact()`, which #142 removes |
| `check_refused_take_marker` | asserts a take refused for a secret leak; nothing can refuse one after #142 |
| `check_secret_floor`, `check_evidence_leak_is_fatal` | #144 arms, unreachable once #142 lands |

Either ordering leaves a slice red on its own — the one thing the split
existed to prevent. So the suite moves once, and the three source slices
behind it are each a deletion in a single file, graded by a suite that no
longer mentions masking.

```
tests/smoke       12,369 -> 8,506 lines
tests/README.md    2,286 -> 1,504 lines
smoke: PASSED — 96 assertions (133 before)
```

## `run_evidence` was rewritten, not deleted — this is the part to review

Per-beat evidence is not a masking feature and survives whole. The **arm
grading it** was mostly about the mask: a byte sweep for thirteen literals,
"renders outside the mask" controls, a card rewritten every 5 ms that had to
refuse capture rather than guess, withheld markup for a value split across
tags, entity forms, and the mask-before-cap ordering. None of those shapes
exists now.

461 lines → 307, and the assertions are **written rather than trimmed**:

- **the page text is real.** Every other check here is an absence — this
  attribute is not in the markup, this script text is not — and an absence
  holds trivially in an empty file. The control requires a string the fixture
  visibly renders to be in the ARIA tree, over a 400-character floor. The
  catalogue's vacuous sweep, and it was genuinely reachable: the old arm got
  this property for free from its mask assertions.
- **the markup serializer drops what was never on screen** — inline `<script>`
  text, `srcdoc`, and value-bearing attributes. This survives #142 for a
  reason unrelated to secrets (an attribute nothing renders was in no frame
  and no still, so serializing it would make evidence the only place it
  exists), and **it had no check of its own**: every old one reached the same
  code through the mask, so deleting the mask would have deleted the only
  thing watching a serializer that still has this job. The take now injects an
  element holding a script and a `srcdoc`, spotlights it, and requires neither
  in the markup.
- **truncation is marked, never silent** on all three text fields.

## Fault injection

None of the three had ever been watched to fail. All three were injected
against the real recorder and the full evidence take:

```
INJECTED: the serializer stops dropping <script>
smoke: FAILED
  - evidence: beat 6's `html` carries inline script text ('SOURCE_ONLY_NEVER_ON_SCREEN').
    It is source, not screen — nobody put it on the page, and evidence describes
    what was on the page.
  - evidence: beat 6's `html` carries a <script> tag ('<script'). …

INJECTED: srcdoc and data-* survive the strip
smoke: FAILED
  - evidence: beat 6's `html` carries a srcdoc document ('SRCDOC_ONLY_NEVER_ON_SCREEN'). …
  - evidence: beat 6's `html` carries the srcdoc attribute ('srcdoc'). …

INJECTED: the page ARIA snapshot comes back empty
smoke: FAILED
  - evidence: beat 2's ARIA tree is 0 characters, under the 400 this grades.
    Every check below is an absence, and an absence holds trivially in an empty file.
```

Each was restored and the suite re-run green after.

## Carve-out

`_StoryboardFailed` is relocated beside `SmokeFailure` **before** the block
holding it is deleted, and the move is asserted in the deletion script
(`count("class _StoryboardFailed") == 1` after). `run_crash_between` uses it
and always did — the placement was incidental, and a deletion pass that trusts
block boundaries takes it. This is #138's carve-out 2.

## Stated limits

- **Every remaining line of masking is now shipped and completely ungraded.**
  This is the direct cost of the re-cut and it is recorded in
  `tests/README.md`'s Known gaps rather than left to be discovered. Between
  this PR and #142/#143/#144, `redact()`, `register_secret()`, the PTY
  scrubber and the evidence masking family are in the package with nothing
  watching them. Nothing depends on those paths — SKILL.md already states that
  the recorder does not defend against a secret reaching the screen (#141).
- **The speech path goes ungraded with it.** The one take that exercised the
  synthesizer, the cache key, the pacing and the ffmpeg audio mix did so
  because `.tts/` was a leak path it had to prove clean. Also in Known gaps.
- **The fixture is untouched.** `?secret=1` still exists and is still
  documented in the fixture table; it goes with #145.
- **`tests/smoke` still has no injection manifest** (#136). The three
  injections above were run by hand and pasted here, which is the state #136
  describes. This slice does not fix it, but it does make what remains smaller.
- **The suite is faster** — 133 assertions to 96, and a whole browser take
  plus two terminal takes gone — but this PR does not measure the wall-clock
  change or touch #61's fast/full split.

## Not demoable

An assertion suite and its documentation. Verifiable only by reading a file or
running the suite.


rogvid:
**Measured on CI:** the smoke job goes **10m43s → 7m34s** (against `main`'s run on #146/#149, same runner class). 96 assertions against 133.

That is the cost side of #61 and #136 starting to unwind — not addressed by this PR, but the number moved and it is worth having on the record before #142/#143/#144 remove the source.

Issue #129, opened 2026-07-27
ticket-queue: search the queue by text


rogvid:
The status filter answers "what is open?" but not "where is that invoice
one?". People scroll. #90 left free-text search out of scope on purpose and
said it was worth its own ticket — this is it.

Add a search box to the queue page. Web only; the CLI keeps `--status` and
gains nothing here.

## Acceptance criteria

1. **A search box above the queue narrows the list as you type.** Typing
   `invoice` leaves only the tickets whose text contains it, case-insensitively,
   with no button to press.
2. **Search and the status filter combine.** With `Open` chosen, typing narrows
   within the open tickets only, and the heading count agrees with what is
   listed — "Support queue (1)", not the count of either filter alone.
3. **Search matches the requester as well as the title.** Typing `mira` finds
   TQ-103, whose title contains no such word; typing part of an address finds
   the tickets from that person.
4. **A search that matches nothing says so, and can be undone.** The queue shows
   the same *"No tickets match this filter."* line the status filter uses, and
   clearing the box restores whatever the status filter alone would show.

## Out of scope

- Searching the ticket body, or anything not on the row.
- Highlighting the matched substring.
- Keeping the search term in the URL. Still a separate ticket, still worth
  doing.

## Demo

This is the feature the **stage 2** reference PR records — issue #64. Unlike
#90 it is one surface, which is deliberate: #96 measured a two-surface take at
61.2s, over the budget the skill states, and stage 2 should be a take that fits
it.

PR #178, opened 2026-07-31
Give a form demo a way to empty a field and press a key (#130)

### LINKED ISSUE #130: No verb clears a field or presses a key, so form demos reach past the recorder into rec.page.keyboard
Found by running the skill on its own use case — the stage 2 reference demo for
#64, which records a search box.

`type_into(selector, text)` puts text into a field, visibly, key by key. Nothing
takes it out again. A search demo empties the box between terms as a matter of
course, and this one had to do it **four times**, so the storyboard reached past
the recorder into Playwright:

```python
def clear_search(rec):
    """Empty the search box the way a person would: select all, backspace."""
    rec.click(SEARCH)
    rec.page.keyboard.press("Control+A")
    rec.page.keyboard.press("Backspace")
```

`rec.page` is right there and it works, which is the problem. What it costs:

- **The keypresses are not beats.** `timeline.json` records the `click` and then
  the next caption; the two keystrokes that actually emptied the field happen in
  the gap between them. An agent reading the beat log cannot tell the box was
  cleared, only that something was clicked and the list changed. Every claim the
  skill makes about the beat log being a faithful account of the take is weaker
  by exactly this much.
- **It bypasses the treatment.** `type_into` types visibly for a reason — a
  field whose contents change between frames with no cursor and no keystrokes
  reads as a jump cut. Backspace here is instant.
- **It bypasses the secret check.** `register_secret` guards captions and the
  text `type_into` is given. `page.keyboard.type()` is not guarded at all, and
  it is the obvious next thing someone reaches for once they have learned that
  `rec.page` is available.

This is not specific to clearing. There is no verb for any key that is not a
character in a string: Enter to submit a form, Escape to dismiss a modal, Tab to
show focus order. All of them are one `rec.page.keyboard` call away, and the
skill neither offers an alternative nor says not to.

## Suggested shape

- `press(key)` — a beat, cursor-free, for a named key. `Enter`, `Escape`,
  `Tab`, `Control+A`.
- `clear(selector)` — a beat that empties a field, visibly. Whether it animates
  backspaces or selects-and-deletes is a taste question; the beat is the point.

Either one alone fixes the reference demo. Both together cover the cases above.

## Acceptance

A storyboard can empty a field and press Enter without touching `rec.page`, and
both show up in `timeline.json` as beats with their key named. The reference
storyboard in `examples/ticket-queue/demos/2026-07-28-queue-search/record.py`
drops its `clear_search` helper.

## Not included

Whether `rec.page` should be discouraged in general. It is a deliberate escape
hatch and this issue does not propose closing it — only that the two things
every form demo needs should not require it.

rogvid:
Closes #130.

Two verbs, `clear(selector)` and `press(key)`, so a form demo stops reaching past
the recorder into `rec.page.keyboard`. The full reasoning is in the commit
message; this body covers only the evidence and the limits.

## Acceptance criterion

A storyboard can empty a field and press a named key, **and both appear in
`timeline.json`**. The second half is the whole point: the reference demo for
#129 reached past the recorder four times, and the beat log recorded a click and
then the next caption, with the two keystrokes that emptied the box happening in
the gap between them.

## What was measured

Six injections, one per claim, each run against the real fixture:

| injection | separation |
|---|---|
| `press` does not press | post-conditions go red |
| `press` does not hold | 0.514 / 0.518 / 0.515 s → 0.013 / 0.013 / 0.012 s |
| `press` loses its `@_beat_verb` | app still changes, **only** the beat-log assertions go red |
| `clear` does not clear | post-conditions go red |
| `clear` loses its pauses | +0.71 s over a click → +0.009 s |
| `clear` loses its `@_beat_verb` | app still changes, **only** the beat-log assertions go red |

The two decorator rows are the filed issue reproduced exactly, and they are the
reason `WEB_EVIDENCE` grew entries keyed on `("clear", "#search")` and
`("press", "Enter")`: every post-condition in this take passes for a storyboard
that drives the page through `rec.page.keyboard` and writes nothing to the log.

`check_form_pacing()` grades the press bar absolutely and the clear bar as what
it costs **over a plain click in the same take**. An absolute bar on clear would
have been satisfied by the cursor glide alone — an assertion that grades nothing.

The first run of the harness reported all six caught for the wrong reason: every
one died on `ModuleNotFoundError` before recording anything, because the driver
had not put the package on `sys.path` the way the suite's own `main()` does.
That is the injection that silently does nothing wearing a red hat, and it is why
the driver now prints the diff it applied and runs an uninjected control first.

## Verified on this branch

- `tests/unit` — 100 tests, OK
- `tests/unit --fault-inject` — all **58** injections caught
- `tests/unit --budget` — SKILL.md 599 lines against a 600 budget
- `tests/smoke` — PASSED (whole suite)
- `tests/smoke --web-only` — `web form verbs hold the frame (3 press beats, clear costs +0.74s over a click)`; the web take is 31 beats, up from 23
- `tests/smoke-inject --self-test` — all 16 registered entries still match
- ruff, ruff format, mypy (12 modules) — clean

## Stated limits

- **The six injections above are not in `tests/smoke-inject`.** They were run by
  an ad-hoc driver, so their evidence lives in this body and not in the
  executable manifest #174 built for exactly this purpose. Six assertions whose
  only record is prose is the thing that manifest exists to prevent. Being fixed
  on this branch before merge.
- **`examples/ticket-queue` still carries its `clear_search` helper.** Dropping
  it means re-recording that take, because its committed `timeline.json` would
  otherwise describe a storyboard that no longer exists.
- **Nothing here discourages `rec.page` in general**, which #130 explicitly did
  not ask for. `rec.page`'s row lost "keyboard shortcuts" from its examples,
  since that is now a verb, and gained "hover-only menus", which is not.
- **`press` takes no selector**, by design — `Tab` *is* the focus demo and
  clicking a target first would destroy it. The safety comes from `type_into`
  and `clear` leaving the caret where they put it. A `press` far from either one
  is the author's responsibility and nothing checks it.

### DIFF
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -251,13 +251,15 @@ exception, or a non-zero exit. Both are
 | `caption(text, ac="AC-3")` / `shot(name, ac="AC-3")` | Tag this beat with the acceptance criterion it is there to demonstrate. Needs `Recorder(criteria={...})`; a tag naming an undeclared criterion is refused. See [reference/review.md](reference/review.md). |
 | `hold(min_s=1.5)` | Keep the current frame up until the current caption's narration finishes (min `min_s`). Use after a spotlight/action so the emphasis rides the whole spoken line instead of flashing. See **Pacing and perception** below. |
 | `move_to` / `click` / `click_fast` / `scroll_to` | Visible cursor motion; `click_fast` for elements that re-render continuously |
-| `type_into(selector, text)` | Click a field and type visibly, key by key — form demos (checkout, login, search) |
+| `type_into(selector, text)` | Click a field and type visibly, key by key — form demos (checkout, login, search). Types at the caret: it appends to what is already there, so `clear()` first to replace it |
+| `clear(selector)` | Empty a field, visibly — click, select what is in it, delete. Its own beat, because emptying a field is something the viewer watches happen |
+| `press(key, hold_s=0.5)` | Press one named key wherever the focus already is — `"Enter"` to submit, `"Escape"` to dismiss, `"Tab"` to move on, `"Control+A"`. Playwright's key names; an unknown one raises. Selector-free on purpose: `Tab` *is* the focus demo and `Escape` acts on whatever is up, and `type_into`/`clear` leave the caret where they put it. Holds `hold_s` so the change is on screen long enough to read |
 | `wait_for(selector)` | Wait for something the app does on its own |
 | `spotlight(selector)` | Ring + enlarge the element the caption discusses; `spotlight()` clears. It eases in *and out* over 250 ms, and the verb waits out its own exit, so the element is back exactly as it was found before the next beat starts (~250 ms per clear on an ordinary take, ~0 under `deterministic=True`, which flattens the transition) |
 | `terminal(cmd)` / `terminal_output(text)` / `terminal_close()` | A *decorative* on-screen terminal card for off-browser actions **inside a web demo** — a prop, not a real shell. To record an actual CLI/TUI use `TerminalRecorder` (below). |
 | `interlude(text, hold=2.8, style=…)` | Bridge a jump; `hold` is how long the card stays before the storyboard moves on. `style="card"` (default) is a full-screen title card, for real time-skips; `style="light"` is a centered label over a soft scrim with the scene still visible, for short transitions. **`interlude("")` fades out whatever is up, whichever style raised it** — the clear takes no `style` and ignores the one it is given. Leave one up and the take says so on stderr and in `content.warnings`; nothing else will notice (see [reference/limits.md](reference/limits.md)). |
 | `stitch(out_dir, [segments])` | Lossless concat of segment recordings into demo.mp4, **and** merge their beat logs into one `timeline.json` / `timeline.md` beside it. `keep_parts=True` leaves each `.seg.mp4` and its `.seg.timeline.*` on disk for a re-stitch |
-| `rec.page` | The live Playwright page — the escape hatch for anything the verbs don't cover (iframes, keyboard shortcuts, drag) |
+| `rec.page` | The live Playwright page — the escape hatch for anything the verbs don't cover (iframes, drag, hover-only menus) |
 
 ## Pacing and perception
 
--- a/skills/demo-video/helpers/demo_recording/content.py
+++ b/skills/demo-video/helpers/demo_recording/content.py
@@ -259,11 +259,13 @@ CONTENT_STATIC_WARN_S = 15.0
 # copied it out of here into a shipped table of beat verbs.
 CONTENT_ACTING_VERBS = frozenset(
     {
+        "clear",
         "click",
         "click_fast",
         "goto",
         "key",
         "move_to",
+        "press",
         "run",
         "scroll_to",
         "send",
--- a/skills/demo-video/helpers/demo_recording/web.py
+++ b/skills/demo-video/helpers/demo_recording/web.py
@@ -155,6 +155,31 @@ window.__demoTerminalHide = () => {
 };
 """
 
+# How long `press()` holds the frame after the key goes down (issue #130).
+#
+# A key press is instantaneous in a way typing is not: the list refilters, the
+# dialog closes, the focus ring jumps, and it is over inside one frame. A verb
+# that returned the moment Playwright did would put the change and the *next*
+# verb's action in the same frame of video, which is the jump cut `type_into`'s
+# per-character delay exists to avoid — the whole reason the storyboard is not
+# just driving Playwright.
+#
+# 0.5 s is a floor with a reason rather than a round number: SKILL.md's
+# "Pacing and perception" puts a saccade at ~200 ms and says a change needs a
+# fixation after it to be recognised rather than merely noticed. It is
+# deliberately *below* the 1.5 s emphasis floor, because a press is a beat in
+# the middle of an interaction and not the emphasis itself — a storyboard that
+# wants the result dwelt on says `hold()` after it, as it does after a click.
+PRESS_HOLD_S = 0.5
+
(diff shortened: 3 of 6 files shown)


rogvid:
## The first stated limit is closed: the six injections are in `tests/smoke-inject`

Registered in 2f907c5. The manifest holds **22 entries** now (16 before), the six
new ones all on `--web-only` — `check_form_pacing` and `check_evidence("web", …)`
are called from `run_web` and nowhere else, so 123 s an entry is genuinely the
cheapest arm that reaches them.

`tests/smoke-inject --arm web` ran the arm **clean first** — that is the
uninjected control, and it is what makes the six reds below attributable to
their breaks rather than to a driver that never got as far as recording, which
is how the first pass of this work went wrong:

```
smoke-inject: 6 injection(s) over 1 arm(s), plus 1 clean baseline(s)

BASE  --web-only passes clean (144s)
```

Then the six, each of which had to go red **and** print the assertion the entry
names in smoke's own failure list. Verbatim:

```
PASS  break: press logs the key it names and never presses it  [137s]
      --web-only, in web.py: self.page.keyboard.press(key)…
      caught: web: the evidence for beat 20 (press 'Enter') does not say ['snapshot 3 of 3', '$119,180', 'Pine & Poplar', 'Cascade Outfitters'] was on screen. A rev
      caught: web: the evidence for beat 20 (press 'Enter') still says ['snapshot 2 of 3', '$134,950'] was on screen, which belongs to an earlier state of the page

PASS  break: press returns in the same frame the key went down in  [133s]
      --web-only, in web.py: self.page.keyboard.press(key)
        self.pause(hold_s)…
      caught: web: the `press` beat for 'Enter' spans 0.014s, under the 0.35s bar — the key went down and the verb returned in the same frame, so the change it caus
      caught: web: the `press` beat for 'Tab' spans 0.014s, under the 0.35s bar — the key went down and the verb returned in the same frame, so the change it caused
      caught: web: the `press` beat for 'Escape' spans 0.012s, under the 0.35s bar — the key went down and the verb returned in the same frame, so the change it cau

PASS  break: press drives the page and writes no beat  [134s]
      --web-only, in web.py: @_beat_verb("press", lambda args, kwargs: args[0] if args else k…
      caught: web: 0 beats are ('press', 'Enter'), so the evidence facts written for it cannot be checked against one
      caught: web: timeline.json holds 0 `press` beat(s), expected 3 — either the storyboard changed or press() stopped recording a beat, which is the whole of what

PASS  break: clear selects the text and never deletes it  [133s]
      --web-only, in web.py: self.page.keyboard.press("Backspace")…
      caught: web: the evidence for beat 18 (clear '#search') does not say ['Ferrari Logistics', 'Redwood Kitchens', 'Blue Ridge Foods'] was on screen. A reviewer h
      caught: web: the evidence for beat 20 (press 'Enter') does not say ['Pine & Poplar', 'Cascade Outfitters'] was on screen. A reviewer holding only this file ca

PASS  break: clear empties the field between two frames  [131s]
      --web-only, in web.py: self.pause(CLEAR_SELECT_HOLD_S)
        self.page.keyboard.press…
      caught: web: the clear('#search') beat spans 0.971s against 0.953s for click('#refresh'), a difference of 0.018s under the 0.4s bar — clear() is doing no more

PASS  break: clear empties the field and writes no beat  [131s]
      --web-only, in web.py: @_beat_verb("clear")…
      caught: web: 0 beats are ('clear', '#search'), so the evidence facts written for it cannot be checked against one
      caught: web: timeline.json does not hold exactly one clear('#search') beat and one click('#refresh') beat (spans None and 0.9489999999999998), so what clear()

All 6 injections were caught. [15.7 min]
```

(The `caught:` lines are cut at 150 characters by the harness's own printer, not
by hand.)

Each red names the assertion it was aimed at, and the two decorator entries name
the beat that went missing by its key — no import error, no missing fixture, no
crash before recording. `tests/smoke-inject --self-test` still passes with all
22 entries: it refuses any entry whose search string does not occur **exactly
once** in the file it targets, in the working tree and again in the staged copy
that actually runs.

### Where each entry is aimed, and why

`check_form_pacing` reads `timeline.json`, so it sees a verb that stopped
logging or stopped holding the frame and is blind to one that logged a beautiful
beat and touched nothing. `check_evidence` reads what the page said when the
verb returned, so it sees exactly the opposite. That is why the three `press`
entries do not all name the same site.

| entry | break | site that must notice |
|---|---|---|
| press logs the key it names and never presses it | `self.page.keyboard.press(key)` deleted | `check_evidence` |
| press returns in the same frame the key went down in | `self.pause(hold_s)` deleted | `check_form_pacing` |
| press drives the page and writes no beat | the `@_beat_verb` line deleted | both |
| clear selects the text and never deletes it | the `Backspace` deleted | `check_evidence` |
| clear empties the field between two frames | both pauses deleted | `check_form_pacing` |
| clear empties the field and writes no beat | the `@_beat_verb` line deleted | both |

### Limits of the six entries themselves

- **The two `check_evidence`-only entries cannot name their beat in
  `must_contain`.** The message interpolates `(press 'Enter')` rather than
  writing it as a literal, and the manifest only accepts expected failures that
  are literals of the named site. What rules out an unrelated red there is the
  clean baseline plus the conjunction of both halves of the per-beat claim, not
  the wording; the entries say so in their comments.
- **No entry asserts that the *other* checks stayed green.** The PR body's
  "app still changes, only the beat-log assertions go red" is not expressible in
  `Injection`, whose verdict is presence-only. The decorator entries pin that
  the beat-log assertions fire, and no more.
- **The manifest is now ~31 min in full**, up from ~16.4 — six entries at 123 s
  plus one 123 s baseline. It is nightly rather than per-push, and `--arm web`
  runs this block alone in ~16 min.
- `tests/README.md`'s "all 15, ~14 min" was already stale before this change and
  is staler now; out of scope here, filed separately.

Issue #200, opened 2026-07-31
check_caption and check_content_healthy are reachable at 28 s now and still have no injection


rogvid:
#197 split the cheap takes out of `--web-only` and `--terminal-only`. While
measuring it, two more of `tests/smoke-inject`'s `ungraded` names turned out to
be reachable from an arm nobody has to think twice about running.

## The measurement

`record_segments` calls `check_caption` three times (the opening line, the
cleared caption and `PROBE_CAPTION`), and `run_segments` reaches
`check_content_healthy` through `check_take`. So both are graded on
`--segments-only`, measured at **28 s** on this box — not only on `--web-only`
(133 s) and `--terminal-only` (187 s), which is what `tests/README.md`'s
roster still implies by listing them beside the pixel checks.

Neither has an entry, and #197 added two entries to that same arm, so the
baseline is already being paid.

## Acceptance

An `Injection(...)` for each, aimed at `--segments-only`:

- `check_caption` — a recorder that stops drawing the caption bar, or draws it
  at zero opacity. It is the skill's headline feature and a take without it
  satisfies every pixel metric in the file.
- `check_content_healthy` — the scored rect wandering outside the app rect it
  was read off. `ContentRect` in `tests/unit` grades the geometry browser-free
  (#195); what has no cheap equivalent is that the rect is *wired in* on the
  stitched path, which is the take this arm records.

Each seen to fail, with the failure output in the pull request, and the
README roster updated to move them off "genuinely ungraded".

## Not included

`check_take`'s artifact half, the narration pair, and the frame-pixel checks
(`_check_frame_captions`, `_check_occlusion`). Those are cost or pixels and
neither is what this issue is about.

rogvid:
Superseded by #241, which carries these two entries plus the six neighbours on the same 29 s arm. Closing #241 closes this.

rogvid:
Superseded by #241, which measured the same arm properly. #200 counted two ungraded checks on `--segments-only`; the AST mapping in #242 found **eight** — `check_take`, `check_beat_frames` and `stitch_segments` carry six more in transitively. Closing #241 closes what this asked for, against a correct count.

Issue #196, opened 2026-07-31
tests/smoke-inject's ungraded list and tests/README.md do not know about tests/unit's coverage


rogvid:
After #195, three of the check functions `tests/smoke-inject` prints as
`ungraded` have their load-bearing claims certified in `tests/unit` instead —
`check_issues`, `check_timeline`, and `check_take`'s picture half via
`check_content_healthy`. Nothing says so where anyone looks.

## The measurement

`tests/smoke-inject --list` still ends with the same 22 names, and then this
line:

> That list is the manifest's boundary, not a to-do list: see tests/README.md
> for which of them are covered by tests/unit at sub-second cost and which are
> genuinely ungraded.

`tests/README.md` predates #195 and does not mention `TakeIssues`, `BeatLog`,
the `ContentRect` geometry assertions, or the `TimelineMd` issue/exit tests. So
the pointer resolves to a document that answers the question wrongly — a reader
following it concludes `check_issues` is graded by nothing, which stopped being
true.

## Acceptance

`tests/README.md` names, per ungraded check function, either the `tests/unit`
class that carries its browser-free claims or "genuinely ungraded", and
`tests/smoke-inject`'s closing line still resolves to something true.

## What this deliberately does not propose

Registering smoke injections for those three. That is the expensive answer and
it is what #173 measured the cost of.

Issue #184, opened 2026-07-31
tests/README.md's smoke-inject entry count and runtime are stale (says 15 entries / ~14 min, now 22 / ~31 min)


rogvid:
`tests/README.md` describes the injection manifest in two places with numbers that no longer hold:

- line 41: `smoke-inject       # proves smoke's assertions can still fail (~14 min, nightly)`
- line 94: `tests/smoke-inject                # all 15, ~14 min`

Measured now, from `tests/smoke-inject --list` on this box:

    22 entries, 6 arms, ~30.6 min of takes

The '15' was already wrong before #178 (the manifest held 16), and the six form-verb entries added on `form-slice` take it to 22 and the full runtime to ~31 min.

Also worth a line while there: the `--arm coverage # one arm's entries (~48 s)` example still holds, but there is now a `--arm web` block that costs ~16 min on its own, which is the one a reader most needs warning about.

Acceptance: the two lines state the count and runtime the manifest actually reports, and nothing else in tests/README.md's 'What the injection manifest covers' section contradicts `--list`.

Deliberately not in scope: making the numbers self-updating. A generated table would drift in a different way and the counts are cheap to re-read.

rogvid:
Folding in #187, filed independently against the same section of the same file while #185 was in flight. Two agents found this drift within an hour of each other, from opposite ends of the manifest.

#187's measurement, which is the more complete one — every figure in `tests/README.md`'s "What the injection manifest covers":

| claim in the README | actual, before #185 | actual, now |
|---|---|---|
| "Sixteen entries" | 22 | **25** |
| per-arm table (5 / 2 / 4 / 2 / 3 = 16) | five arms, 22 | **six arms, 25** |
| "23 of `tests/smoke`'s 32 check functions have no entry" | 22 of 33 | **22 of 35** |
| "16.3 and 16.4 minutes" end to end | measured at 16 entries | **not re-measured (~31 min)** |
| line 41 `~14 min, nightly` | already wrong | ~31 min |
| line 94 `all 15, ~14 min` | already wrong | 25 |

`tests/smoke-inject`'s own docstring was corrected in #185; the README was not.

**Why this is worth more than the usual stale number.** This section is the one a reader consults to find out *what the manifest does not cover* — it is the honesty surface for the whole apparatus. A coverage claim that reads better than the truth ("23 of 32 uncovered" when it is 22 of 35) is the failure mode the manifest exists to prevent, committed by the document describing it.

[...]

Issue #186, opened 2026-07-31
A deterministic take's opening stills are not byte-reproducible: the cursor dot's position is a race


rogvid:
Split out of #185, which fixed this inside `tests/smoke` only.

## The measurement

`Recorder(deterministic=True)` announces "Re-recording reproduces the take."
For stills that is not true of a storyboard that never moves the pointer.

The recorder draws `#__demo_cursor` wherever the last `mousemove` landed. A
storyboard that never moves the pointer still gets one: Chromium dispatches a
synthetic `mousemove` at its last known pointer position — `(0, 0)` on a fresh
page — once layout settles, and the handler puts an 18 px dot in the top-left
corner. Whether that arrives before the first `shot()` depends on load.

Measured on a 16-core box under 14-way CPU load, twelve takes of the entropy
storyboard:

| dot at | takes |
|---|---|
| the viewport origin | 11 |
| off-screen, where the stylesheet parks it (`top`/`left: -40px`) | 1 |

Two stills that disagree differ by 69 of 921 600 pixels, all inside x 0-8,
y 0-8. That is what #185 turned out to be, after three CI runs and a bisect of
`main`.

## Why it is not fixed there

#185 fixed it in the harness: `tests/smoke`'s determinism storyboard now parks
the pointer explicitly and asserts it stayed parked. That makes the *test*
honest. It does not help a user, whose storyboard has no reason to move the
pointer before a `shot()`, and whose first still is therefore a coin flip
between two pictures.

## Acceptance

A take recorded twice with `deterministic=True`, from a storyboard that never
moves the pointer, produces byte-identical stills — and the arm that proves it
does not itself place the pointer. Options worth weighing, none picked here:

- park the cursor deterministically at recorder start-up, so the dot has one
  position until a verb moves it;
- keep the dot hidden until the first pointer verb, so a take that uses none
  has no dot at all;
- treat the synthetic move as not-a-move — ignore an untrusted `mousemove`
  that arrives before any verb has run.

The first two change what a recording looks like, so by CLAUDE.md's test they
want a demo.

## Not included

The still comparison and the diagnostic printed beside it. Both landed with
#185.

PR #201, opened 2026-07-31
Make the expensive arms injectable, and the coverage claims true after

### LINKED ISSUE #193: tests/smoke cannot see the cursor race the product now fixes: no arm records a take that never moves the pointer
Split out of #186, which fixed the defect in the recorder (`skills/demo-video/helpers/demo_recording/web.py`) and graded the *rule* in `tests/unit`.

## What is missing

The take-level claim — **two takes of a storyboard that never moves the pointer produce byte-identical stills** — is the acceptance criterion of #186, and no suite grades it.

- `tests/unit` grades the predicate (`CursorMotion`): the guard is read out of the shipped overlay and evaluated against event shapes measured off Chromium. It cannot see a *browser* that changed — a settle event that started reporting movement would reopen #186 with the unit suite green.
- `tests/smoke`'s determinism arm parks the pointer at `PARKED_POINTER = (60, 640)` before anything is photographed (#188), so it is structurally unable to record the case #186 is about. `check_parked_pointer` asserts the dot is *at* the park, which passes either way.

So the fix shipped on a 24-take measurement made by hand, outside the suite:

| | takes | dot rect (live page) | distinct still hashes |
|---|---|---|---|
| before | 24 | `(0,0)` x12, off-screen x12 | 2 |
| after | 24 | off-screen x24 | 1 |

(16-core box, 14-way CPU load, `deterministic=True`, `tests/fixture`, dot read with `getBoundingClientRect`, stills hashed as bytes.)

## Acceptance

One arm records the pointer-free case at least twice and requires the stills to be byte-identical, **without placing the pointer itself** — the existing park is what makes the current arm blind, so the new take cannot reuse it. Plus a check, beside `check_parked_pointer` and read the same way (a rect off the live page, never the recorder's inline `left`/`top`), that the dot is where the stylesheet parks it (centre near `(-40, -40)`) in a take no verb moved the pointer in.

Then a `tests/smoke-inject` entry that removes the guard —

```
if (e.movementX === 0 && e.movementY === 0) return;
```

— from `web.py` and requires that check to fail. Note what the injection has to prove: with the guard gone the *still comparison* alone notices roughly one run in two on this box and rather less on a quiet one, so the entry must name the position check, not the hashes. An intermittent assertion is how #185 cost three CI runs.

## Not included

The rule itself and its three injections; both landed with #186.

### LINKED ISSUE #197: Split the cheap takes out of --web-only and --terminal-only so their checks are injectable
The arm-splitting half of #173. #195 closed #173 by pulling the *browser-free*
claims of `check_issues`, `check_timeline` and `check_take`'s picture half down
into `tests/unit`, where an injection costs 0.2 s. What it did not do — and
deliberately, the content arm was its whole scope — is make the smoke-side
assertions themselves reachable at a price anyone will pay.

## What remains, from #173's own checklist

- [ ] `--issues-only` — `run_web_problems` + `run_terminal_problems`, which is
      what `check_issues` needs on a real page and a real command. Both are
      short takes against a page that throws and a command that exits 3.
- [ ] `--timeline-only`, or reuse of an existing cheap take, for
      `check_timeline` / `check_beat_frames`: they grade the beat log and the
      review frames of *a* take, not specifically the long web one.
- [ ] an entry aimed at `check_evidence` from `--evidence-only`, which already
      exists at 7 s. That one is a manifest gap rather than a cost gap and is
      the cheapest thing here to close.

`--strict-only` (added in #136) is the worked example: two takes recording in
13 s between them, previously reachable only from the 123 s and 186 s arms.

Each `run_phases` split is a couple of lines and must keep the full-suite
ordering unchanged (`only in (None, ...)`).

## Why it is still worth doing after #195

`tests/unit` grades what the recorder *writes down*. It does not grade that
Chromium fires `console` for a `console.error`, that a refused connection
arrives as `requestfailed` rather than as a response, or that the beat which
was open is the beat that was on screen. Those are smoke's, and today an
injection against them costs 123-186 s plus the same again for the baseline.

## What this deliberately does not propose

Adding entries at the current arm prices. A manifest nobody can afford to run
is worth less than a smaller one that runs nightly.

rogvid:
Closes #197 and #196. One slice: make the expensive arms injectable, then make the coverage claims true afterwards.

> **#193 is no longer closed by this PR.** The `--pointer-only` arm it asked for was built, measured and removed before merge — Chromium dispatches the event it needs only when the page's `load` fires inside 22 ms (7 of 7 takes under that bar, 0 of 9 over it), which no storyboard controls. **#210** carries the measurement and the gap. The table below still describes the arm; the later comments describe its removal.
> 
> The commit `Make the expensive arms injectable...` still carries a `Fixes #193` trailer from before that decision. If this is merged as a merge commit rather than a squash, that trailer will close #193 — worth dropping or closing #193 as superseded by #210 by hand.

## The arms, measured on this box

| arm | before | after | what it now reaches |
|---|---|---|---|
| `--evidence-only` | 7 s | **6 s** | `check_evidence` (was `--web-only`/`--terminal-only` only) |
| `--pointer-only` | — | **15 s** | `check_resting_pointer` (new; no arm could record this) |
| `--issues-only` | — | **26 s** | `check_issues` (was 123 s / 186 s) |
| `--segments-only` | 29 s | **28 s** | `check_timeline`, `check_beat_frames` — it already did; nothing was aimed at it |
| `--web-only` | 123 s | **133 s** | unchanged; every split is additive |
| `--terminal-only` | 186 s | **187 s** | unchanged |
| the whole suite | 622 s *(documented)* | **427 s** | passes, `run_phases` ordering unchanged |

The 622 s was never checked and was wrong: `--web-only` and `--terminal-only` are disjoint and account for 320 of the 427 s, with the determinism, segments, failure and pointer arms making up the rest. `tests/README.md` now carries 427 s and says why.

Manifest: **25 entries over 7 arms (~31.9 min) → 31 over 11 (~35.4 min)**. Ungraded check functions: **22 of 35 → 19 of 36**.

## Every new entry, seen to fail

**#193 — `check_resting_pointer`, `--pointer-only`.** Removing `if (e.movementX === 0 && e.movementY === 0) return;` from `web.py`. The **whole failure list** of the injected arm, not only the matched lines:

```
smoke: FAILED
  - resting-0: in resting-0, no verb moved the pointer and the cursor dot is centred at (0, 0), not at the (-40, -40) the recorder's own stylesheet parks it at — something other than the storyboard is placing the cursor, which is issue #186: a mousemove reporting no movement is Chromium recomputing hover at the position the pointer is already at, and following it makes the dot's position a race
  - resting-1: in resting-1, ... (identical)
  - resting-2: in resting-2, ... (identical)
```

**That is the entire list — the still-hash comparison stayed silent.** All three takes moved to the origin, so the stills came out byte-identical and #193's byte claim passed against a recorder with the defect in it. That is the measurement the issue asked for, and it is why the entry names the position check.

**#197 — `check_issues`, `--issues-only`, two entries.**

```
PASS  break: an event is blamed on the most recently started beat  [26s]
      caught: web-problems: the console_error 'fixture: between beats' fired between beats, with none open, but is attributed to beat 1 ('wait_for') — the attributi

PASS  break: a command's exit status lands on the newest run() rather than the oldest  [26s]
      caught: terminal-problems: run() beats logged exit codes {'echo ok': 0, '(exit 3)': 3, 'sleep 1': 9, '(exit 9)': 0}, the storyboard's commands exit {'echo ok'
```

**#197 — `check_timeline` and `check_beat_frames`, `--segments-only`, two entries.**

```
PASS  break: the merge renumbers each beat's position within its own segment  [28s]
      caught: segments: beat 6 ('interlude') carries segment_index 6, expected 0 — it is the beat's position within its own segment, and unlike `index` a merge must
      caught: segments: beat 7 ('goto') carries segment_index 7, expected 1 — ...

PASS  break: every review frame is cut at the start of its beat, not its midpoint  [28s]
      caught: segments: beat-00.png was taken at 0.406s, but beat 0 runs 0.406-0.963s and its midpoint is 0.684s — the frame is not the moment the sheet numbers it
      caught: segments: beat-02.png was taken at 0.991s, but beat 2 runs 0.991-3.003s and its midpoint is 1.997s — ...
```

**#197 — `check_evidence`, `--evidence-only`, one entry.**

```
PASS  break: every beat's evidence describes the first beat's page  [6s]
      caught: evidence: the evidence for beat 5 (spotlight '#smoke-source') does not say ['this element renders one line'] was on screen. A reviewer holding only th
      caught: evidence: the evidence for beat 7 (spotlight '#smoke-bloat') does not say ['bloat row 0'] was on screen. A reviewer holding only this file cannot stat
```

**#196 — the README roster, two new `--self-test` guards.** Each is applied to a copy of the README, required to land (`re.subn` hits) before it is required to be caught:

```
CAUGHT   a README whose ungraded roster lost a row
CAUGHT   a README whose roster answers a gap with nothing in particular
```

## Two entries that graded nothing, and how that came out

The first entry aimed at `check_issues`' attribution named `if not self._in_beat or not self._beats:`. It passed against a broken recorder. The second named the other guard, `if t_start > self._pumped_at + ATTRIBUTION_SLACK_S:`. It passed too. Instrumenting `_attributed_beat` on the real take showed why: the between-beats console error is delivered **inside the `page.evaluate` that raises it** — `in_beat=False`, last beat `wait_for` at `t_start 0.934`, `_pumped_at 0.383` — so *both* guards refuse it independently and removing either alone is a behavioural no-op. The entry that landed replaces the whole body with the naive `self._beats[-1]`.

Neither miss would have been found by reading, and both would have shipped as "proof" that the assertion grades something.

## What #196 changed, and the property it preserves

`ungraded` is printed per run and has always meant "this manifest aims nothing at it" — never "nothing grades it". After #195 that difference was most of the list, and the closing line pointed at a `tests/README.md` written before it.

The README now carries a row per ungraded check function naming either the `tests/unit` class that carries its browser-free claims (`ContentRect`, `TakeIssues`, `VerbClassification`, `OpeningWarning`, `CaptionTruth`) or "genuinely ungraded", with what is given up stated in each case. `readme_ungraded_roster()` reads it **as text** and compares against `ungraded()` — `INJECTIONS` against the AST of `tests/smoke` — **both ways**: a name missing from the roster is a reader sent to a table their function is not in, and a name in it that has an entry is #184's mistake again. No number is taken from the README and then checked against the README.

`--self-test` now compares 7 figures, 2 tables and the roster.

## Stated limits

- **The pointer arm can refuse rather than fail.** With the guard removed the dot moves in ~14 of 16 takes; the rest receive no `mousemove` at all. Three takes make a miss ~1 in 500, and the run that would have missed it is the run whose control guard reports "not one of the 3 takes saw a mousemove reporting no movement" — an honest refusal, never a pass. The cost is a ~0.2% chance the arm goes red for its premise.
- **`--web-only`'s 123 s is stale** (re-measured 133 s). Left as written rather than churned on a single reading; the README says so.
- **`check_caption` and `check_content_healthy` are now reachable at 28 s** via `--segments-only` and still have no entry — filed as #200 with the measurement.
- The three pre-existing `check_evidence` entries were not all re-run. One was, as a regression check on the signature change: `clear selects the text and never deletes it`, PASS, 133 s.

## Verification

Run: `./tests/unit` (141 OK), `./tests/ci-unit` (49 OK), `./tests/lint`, `./tests/smoke-inject --self-test`, `./tests/smoke` **whole suite, PASSED, 427 s**, and clean + injected runs of `--pointer-only`, `--issues-only`, `--segments-only` and `--evidence-only`, plus clean `--web-only` (133 s) and `--terminal-only` (187 s).

Not run: the whole `tests/smoke-inject` manifest (~35 min), and the entries on `--content-only`, `--overlay-only`, `--failure-only`, `--coverage-only`, `--strict-only` and `--determinism-only` — none of those arms or their entries is touched here.

No demo video: nothing in this change is watchable. It is test harness and honesty surface, which is the right-hand column of CLAUDE.md's table.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -49,10 +49,9 @@ tests/
 [#136]: https://github.com/rogvid/skills/issues/136
 
 Takes: `web/` and `terminal/` (the two media), the determinism pair, the
-three `resting-*` takes that never place the pointer, the problem takes,
-`segments/` — one demo recorded in two parts and joined with `stitch()` — and
-`evidence/`, a few seconds against what a beat's page text carries and what it
-caps.
+problem takes, `segments/` — one demo recorded in two parts and joined with
+`stitch()` — and `evidence/`, a few seconds against what a beat's page text
+carries and what it caps.
 
 ## Running it
 
@@ -97,8 +96,6 @@ tests/smoke --strict-only         # just the two takes strict=True must refuse
 tests/smoke --issues-only         # just the broken page and the failing
                                   #   commands, which is what check_issues
                                   #   grades (issue #197)
-tests/smoke --pointer-only        # just the takes that never place the
-                                  #   pointer (issue #193)
 tests/smoke --out-dir /tmp/smoke  # keep the recordings at a known path
 tests/smoke --keep                # keep the temp dir even when it passes
 ```
@@ -110,7 +107,7 @@ grades, the manifest that proves it can still fail:
 tests/smoke-inject --self-test    # the harness's own guards (instant)
 tests/smoke-inject --list         # every entry, its arm and what it costs
 tests/smoke-inject --arm coverage # one arm's entries (~48 s)
-tests/smoke-inject                # all 31 entries, ~35 min
+tests/smoke-inject                # all 30 entries, ~35 min
 ```
 
 Those three figures, and every number in **The injection manifest** below, are
@@ -736,44 +733,6 @@ consistently. The take reads the line and requires `worker <frozen epoch>`.
 All three takes also run `strict=True` (issue #3's machinery), which is what
 would catch the blob shim breaking worker *loading* rather than its clock.
 
-**And the take the park makes impossible** — `tests/smoke --pointer-only`,
-three recordings of the same storyboard with the pointer left exactly where it
-was ([#193](https://github.com/rogvid/skills/issues/193)). The determinism
-takes above park the mouse at `(60, 640)` before anything is photographed,
-which is what makes them reproduce — and also what makes them structurally
-unable to record the case
-[#186](https://github.com/rogvid/skills/issues/186) fixed: a storyboard that
-never moves the pointer, where the recorder's cursor dot is placed by
-Chromium's load-time hover event or by nothing at all. `tests/unit` grades the
-*rule* (`CursorMotion` evaluates the shipped guard against event shapes
-measured off Chromium) and cannot see a browser that changed its mind about
-reporting movement. This arm can.
-
-Three things, in this order:
-
-1. **the control.** At least one of the three takes has to have received a
-   `mousemove` reporting no movement — counted by a listener installed as an
-   init script, because the only one this storyboard gets is dispatched while
-   the page is loading and a listener added after `goto()` returns counts zero
-   of them (measured, 0 events over 10 takes against exactly one, at `(0, 0)`
-   with both deltas 0, when installed before). Without it "the dot is parked"
-   is also true of a take the browser never sent an event to, and the arm
-   would be grading the machine.
-2. **the dot is where the recorder's own stylesheet parks it** — centre
-   `(-40, -40)`, read as a rect off the live page through
-   `getBoundingClientRect`, never as the inline `left`/`top` the recorder's
-   handler writes.
-3. **the three stills are byte-identical**, which is #186's acceptance
-   sentence.
-
-The `tests/smoke-inject` entry names the *second* of those, not the third, and
-that is the whole reason the position check exists. With the guard removed the
-dot lands at the origin in about seven takes in eight — measured, 14 of 16 —
-and stays parked in the rest, both stable within a take. So the byte comparison
-only notices when two takes happened to disagree, while the position check
-notices any take that moved. An intermittent assertion is what
-[#185](https://github.com/rogvid/skills/issues/185) cost three CI runs.
-
 **Segments and the merge** — a demo recorded in two parts and joined by
 `stitch()` ([#7](https://github.com/rogvid/skills/issues/7)). `segments/`
 records the storyboard `SKILL.md` prescribes for a real time-skip: part one,
@@ -1370,7 +1329,6 @@ are what changed that, and they are now load-bearing rather than a convenience.
 | `--coverage-only` | 8 s |
 | `--narration-only` | 8 s |
 | `--strict-only` | 13 s |
-| `--pointer-only` | 15 s |
 | `--determinism-only` | 19 s |
 | `--issues-only` | 26 s |
 | `--polish-only` | 26 s |
@@ -1381,6 +1339,11 @@ are what changed that, and they are now load-bearing rather than a convenience.
 | `--content-only` | 148 s |
 | `--terminal-only` | 186 s |
 | the whole suite | 427 s |
+
+That whole-suite reading includes a 15 s arm the same branch dropped before
+merge (#210), so the tree as it stands is nearer 410 s. It is left as measured
+rather than corrected by subtraction — a number nobody has read off a stopwatch
+is how this row came to say 622 s.
 
 The per-arm figures are the calibration this manifest was built on. The
 whole-suite row is a fresh reading taken while
@@ -1438,7 +1401,7 @@ easy to break. This is what `tests/smoke-inject --list` reports today, quoted
 rather than remembered:
 
 ```
-31 entries, 11 arms, ~35.4 min of takes
+30 entries, 10 arms, ~34.9 min of takes
 ```
 
 That figure is `--list`'s **estimate** — each entry's arm from the table above,
@@ -1460,7 +1423,6 @@ paragraph whose job is to say what this manifest does not cover.
 | `--coverage-only` | 5 | the report flatters the storyboard: nothing reported unclaimed, a claim pointing at the wrong beat or forgetting its segment, an undeclared tag accepted, the finding dropped from `timeline.md` |
 | `--strict-only` | 2 | a take that should have been refused passes, or the refusal never says which beat caused it |
 | `--determinism-only` | 3 | a re-recording stops reproducing: the pointer parked wherever the race left it, a frozen clock that freezes at the wall time, an animation still moving when the still is taken |
-| `--pointer-only` | 1 | the recorder's cursor dot goes back to chasing the `mousemove` Chromium fires at load ([#186](https://github.com/rogvid/skills/issues/186)) — aimed at the dot's *position*, never at the still hashes beside it, because with the guard gone the two takes agree about as often as they disagree |
 | `--issues-only` | 2 | what the recorder saw behind the pixels stops being true: a problem that fired with no beat open acquires a confident beat index and caption, or a command's exit status lands on the wrong `run()` beat and a failure is recorded as a success |
 | `--segments-only` | 2 | the merge renumbers `segment_index`, so `(segment, segment_index)` stops naming the same beat across a stitch ([#22](https://github.com/rogvid/skills/issues/22)); or every review frame is cut at its beat's start instead of its midpoint, and the sheet is caption fade-ins |
 | `--evidence-only` | 1 | one capture of the page is stamped onto every beat, so what a beat's evidence describes is not what that beat showed — [#9](https://github.com/rogvid/skills/issues/9)'s acceptance criterion, on a 7 s arm instead of a 123 s one |
@@ -1471,7 +1433,7 @@ paragraph whose job is to say what this manifest does not cover.
 
 ### What it does **not** cover
 
-- **19 of `tests/smoke`'s 36 check functions have no entry**, and the harness
+- **19 of `tests/smoke`'s 35 check functions have no entry**, and the harness
   prints every one of them as `ungraded` at the end of a run rather than
   leaving the boundary to somebody's memory.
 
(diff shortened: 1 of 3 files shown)


rogvid:
## The rebase failure: case 1, established rather than inferred

**The load-time event still arrives on this build.** Chromium 151.0.7922.34 (playwright 1.61), the third row of #202's table. Measured two ways:

| how | takes | received a `mousemove` |
|---|---|---|
| listener installed as an init script, read at 0.3–6.0 s | 16 | 16 |
| `--pointer-only`, five clean runs, counted by the arm itself | 15 | **13** |

So the arm can and does grade #202's rule — and delivery is per-take racy at ~13 in 15. Three takes miss together about **1 run in 500**, and the rebase hit exactly that: all three lists empty. The whole-suite run after this change got 1 of 3, the same phenomenon closer to the edge.

Not case 2, and the difference matters: the control does not need relaxing, it needs a bigger sample. That is **#205**, kept out of this PR because `RESTING_TAKES` moves the arm's cost, which is written down in `tests/smoke-inject` and `tests/README.md` — both out of scope here.

## What changed in `tests/smoke`

The control was asking `movementX === 0 && movementY === 0` — **#194's rule, still living in the harness after #202 took it out of the product**, and wrong here for exactly the reason #202 gives: the delta for a page's first mouse event is a build detail (147 reports 0, 0 for a park at (60, 640); 136 and 151 report 60, 640). On a build reporting a non-zero delta for the settle event it would have refused a take that *did* receive one — a false "cannot grade". It now looks for a `mousemove` repeating `CURSOR_SEED = (0, 0)`, the position the overlay seeds its remembered pointer with, hardcoded rather than imported so it cannot agree with a recorder that seeded it somewhere no event lands.

The one refusal became two, both still failures:

- **saw nothing** — "this arm could not grade the rule it exists for", with the delivery rate and a pointer to #205, so a reader knows whether to re-run or to investigate.
- **saw something that is not a repeat** — the events it got are not the ones this arm is about.

## Both seen to fire, each break required to match exactly once

On takes that **really did receive the event** — the probe logs it one pixel across, so `resting-0` and `resting-2` carry a genuine event and no logged position repeats the seed:

```
break 'wrong-spot' matched exactly once and landed
smoke: FAILED
  - resting-pointer: the takes saw mousemove(s) {'resting-0': [[1, 0, 0, 0]], 'resting-1': [], 'resting-2': [[1, 0, 0, 0]]}, and not one of them repeats the (0, 0) the recorder's overlay seeds its remembered pointer with — so nothing here exercised the guard that drops a move to a position the pointer was already at, which is the rule (#202) this arm grades. A take whose only events are real moves says nothing about it.
```

And the no-event branch, listener never registered:

```
break 'saw-none' matched exactly once and landed
smoke: FAILED
  - resting-pointer: none of the 3 takes saw a mousemove at all, so this arm could not grade the rule it exists for — the dot being parked is equally true of a page the browser never sent an event to. That is the machine, not the recorder: Chromium dispatches this event on a load and 2 takes in 15 do not get one on the build this was measured on. Re-run --pointer-only; if it keeps happening the build has stopped sending it and #205 is where that is recorded.
```

The retargeted registered entry still catches, against the positional rule:

```
BASE  --pointer-only passes clean (15s)
PASS  break: the cursor dot follows a mousemove that reports no movement  [15s]
      --pointer-only, in web.py: if (e.clientX === atX && e.clientY === atY) return;…
      caught: resting-0: ... the cursor dot is centred at (0, 0), not at the (-40, -40) ...
```

## The whole suite does not pass, and it is not this branch

`./tests/smoke` is red on two lines in the **`web`** take. `--pointer-only` passed in every run, including the failing ones.

```
- web: the closing caption could not be timed against demo.mp4 — the caption arrives 0.36s into a 1.2s run-up, leaving no frames before its fade to use as a baseline — the video has slid further under the beat log than the search window allows for (see issue #18)
- web: beat-23.png should show the caption 'Keys, not just clicks.' and its caption band is only 3.26 from the take's caption-off baseline — beat-27.png, which should show no caption at all, is 2.91 from it. The two are 0.34 apart, under the 6.0 floor
```

Reproduced on a staged `origin/main` (8feab92) with this branch absent — four `--web-only` runs, **3 failed**, one passed, at load averages 1.0–3.6. One `main` run also produced `the last beat ends at 30.607s, past the end of a 30.36s demo.mp4`. Filed as **#209** with the measurement; the failure text points at #18, which is closed.

## Verification

`./tests/smoke --pointer-only` clean ×6 and injected ×3 (registered entry plus both control breaks), `./tests/smoke-inject --self-test`, `./tests/unit` (141 OK), `./tests/ci-unit` (49 OK), `./tests/lint`, `./tests/smoke` whole suite ×2 — pointer arm green in both, suite red on #209.

Issues filed: **#205** (the control's error bar), **#209** (the web arm on `main`).

rogvid:
## The load hypothesis is wrong, and the real variable makes the arm unfixable

**Idle vs loaded, this box, Chromium 151.0.7922.34** — the comparison you asked for:

| | takes | received the event |
|---|---|---|
| `--pointer-only`, idle (load 0.7–1.9) | 9 | 8 |
| `--pointer-only`, 12-way CPU load (load 11.1) | 9 | 7 |

Load is not it. I could not reproduce 0/12 idle at any load, on `4109159`, on a tree that already contains `d522548`.

**The probe was already an init script** — `page.add_init_script`, before the `goto`. I instrumented when it actually attaches: `readyState: 'loading'`, **7.2–8.4 ms into document time, in 16 of 16 takes**. There is no earlier installation. Your 16/16 and my 13/15 were not two mechanisms; they were the same one sampled on different days.

**The variable is the page's own `load` timing, and the separation is total.** Recording the `load` event and the mousemove in the same takes:

| `load` fired at | takes | received the event |
|---|---|---|
| ≤ 22 ms | 7 | **7** |
| ≥ 38 ms | 9 | **0** |

No take landed in between. Where it arrives, it arrives 27–39 ms after `load`; where it does not, `load` fires at the moment it would have been dispatched. Sample rows:

```
take 1: listener at 8.1ms readyState='loading' marks=[['DOMContentLoaded', 11], ['load', 12]] moves=[[0, 0, 0, 0, 39]]
take 2: listener at 7.7ms readyState='loading' marks=[['DOMContentLoaded', 10], ['load', 39]] moves=[]
```

`tests/fixture/index.html` pulls **no subresources at all** and still loads in 11 ms or 40 ms from one run to the next, so no storyboard controls this. A second navigation does not help — `web.py`'s own header records the hover-recompute move is sent for the first document only. Rates across pages, all idle: `/?entropy=1` 5/8, `/` 2/8. Your box: 0/12.

## So: option 3. The arm is removed.

Nothing in `tests/smoke` can make Chromium dispatch an event it did not send, and a permanently-red arm trains people to ignore a red run. Removed in `ea51c5f`, with the gap recorded in **#210** (full measurement, plus the one route that might still work: two real `page.mouse.move` calls to the *same* position, which exercises the same guard with an event the harness causes, at the cost of no longer being #186's original sentence).

The removal necessarily reaches past `tests/smoke` — the entry names `check_resting_pointer` as its site, and `--self-test` compares the arm tables and `--list` figures in `tests/README.md` against the manifest. All three moved together: **30 entries over 10 arms, ~34.9 min, 19 of 35 check functions ungraded**. The harness refused the stale entry on the way, which is what it is for:

```
STALE: 'the cursor dot follows a mousemove that reports no movement':
names site 'check_resting_pointer', which is not a function in tests/smoke
```

`tests/README.md` gains a **Known gaps** entry with the numbers, so the sentence #186 and #202 were both accepted on is openly absent rather than quietly so. The whole-suite row keeps its measured 427 s with a note that the reading included this arm — corrected by subtraction is how it came to say 622 s.

The two-branch refusal is gone with the arm. It did its job: it is what surfaced this instead of letting the arm read green on a fast runner.

## What #201 still closes

#197 and #196, unchanged and unaffected — `--issues-only`, `--segments-only`, `--evidence-only`, the five entries against them, and the ungraded roster. **#193 is no longer closed by this PR**; #210 supersedes it with the measurement, and #193 should be closed as superseded or left open pointing at it — your call.

## Verification

`./tests/smoke --issues-only` ×3 clean, `./tests/smoke-inject --arm evidence` (clean + injected), `./tests/smoke-inject --self-test`, `./tests/unit` (141 OK), `./tests/ci-unit` (49 OK), `./tests/lint`. Whole suite not run, per your instruction — #209 is separate.

Issue #214, opened 2026-08-01
The beat right after a click-driven navigation still reports the caption the load took away


rogvid:
Follow-up to #179, filed by the PR that closed it. Measured, not suspected.

## Measurement

A probe drove a real Chromium against a fixture with a link, a GET form and a
`location.href` button, with the shipped `Recorder._watch_page` subscription on
the page, and recorded which page events had been delivered by the time the
driving Playwright call returned. Identical on **Chromium 136.0.7103.25
(Playwright 1.52), 147.0.7727.15 (1.59) and 151.0.7922.34 (1.62)**:

| what the storyboard did | delivered when the call returned | rest |
|---|---|---|
| `goto("/b.html")` | `framenavigated`, `domcontentloaded`, `load` | — |
| `click("#link")` | `framenavigated` | `domcontentloaded`, `load` |
| `click("#submit")` (GET form) | `framenavigated` | `domcontentloaded`, `load` |
| a button assigning `location.href` | `framenavigated` | `domcontentloaded`, `load` |

`goto()` waits for the load (`wait_for_load_state("networkidle")`), so
`_note_document_replaced` has already run when it returns. A click does not:
the document event arrives during whichever Playwright call comes next.

A beat stamps `caption` when it **opens** (`_beat`, `core.py`), before its verb
touches the browser. So the first beat after a click-driven navigation carries
the line that is no longer on screen — exactly the artifact #134 exists to
stop, one beat wide instead of the whole rest of the take.

`tests/unit`s `MeasuredNavigations.test_a_load_the_verb_did_not_wait_for_
reaches_the_log_one_beat_late` pins this as it is today: `[STALE, ""]`, so
closing the gap is a deliberate edit rather than a surprise.

## Acceptance

A storyboard of the shape `caption(...)`, `click("#link")`, `pause(3)` records
no beat carrying the caption after the click. The obvious lever is pumping page
events at beat *open* rather than only inside `_idle`, which costs one
`evaluate("0")` round trip per beat — measure it before taking it.

## Not included

Anything about `goto()`, which was measured clean on all three builds.

Issue #202, opened 2026-07-31
A storyboard that parks the pointer with page.mouse.move records no cursor at all on some Chromium builds


rogvid:
A storyboard that drives the pointer through `rec.page.mouse.move` — the escape
hatch `SKILL.md` documents, and what `tests/smoke`'s determinism arm uses —
records **no cursor at all** on some Chromium builds. `tests/smoke
--determinism-only` fails three ways, one per take:

```
smoke: FAILED
  - determinism: in the determinism take, the cursor dot is centred at (-40, -40), not at the (60, 640) this storyboard parked the pointer at — its position is being set by something other than the storyboard, and whatever that is decides whether two takes of this storyboard produce the same pixels
  - determinism: in the determinism take, the cursor dot is centred at (-40, -40), ...
  - default: in the default take, the cursor dot is centred at (-40, -40), ...
```

Bisected to 2b351dd (#194, fixing #186), which made the overlay ignore a
`mousemove` reporting no movement:

```js
if (e.movementX === 0 && e.movementY === 0) return;
```

## Mechanism

`movementX`/`movementY` is the delta from the previous mouse event's position.
The **first** mouse event in a page has no previous event, and what Chromium
substitutes is a build detail. Measured against `tests/fixture`, park at
(60, 640) with a raw `page.mouse.move` right after `goto`:

| Chromium (playwright) | park's `movementX/Y` | dot, 3 takes |
|---|---|---|
| 136 (1.52) | 60, 640 — measured from the origin | (60, 640) ×3 |
| 147 (1.59) | **0, 0** — measured from the event's own position | **(-40, -40) ×3**, off screen |
| 151 (1.61) | 60, 640, after a settle event at (0, 0) seeds the position | (60, 640) ×3 |

On 147 the storyboard's own park is indistinguishable from the load-time hover
event #186 is about, so the guard throws it away and the dot never leaves the
off-screen position the stylesheet parks it at.

A second symptom, and this one bites on **every** build including CI's: park,
navigate, park again at the same coordinates. The repeat move reports no
movement, and the new document's dot is a fresh element nothing else will
place — off screen in 4 takes of 4 on Chromium 151.

## Why CI never saw it

CI installs Chromium through `uv run --with playwright playwright install`,
i.e. whatever the newest Playwright ships — `chromium-headless-shell v1234`
(151.0.7922.34) in the run of 2026-07-31. That is one of the builds that
measures the first delta from the origin, so the park reports movement and
[...]

Issue #180, opened 2026-07-31
A caption lost to a page load is silent — the beat log now says nothing where a line used to be


rogvid:
Follow-up to #134, filed by the PR that fixed it. #134 deliberately excluded
this: "Warning the author that a caption was lost. Arguably right, and #114
covers the adjacent case for terminal takes; this issue is only about the beat
log reporting what is true."

## What the fix changed

`Recorder._note_document_replaced` clears `self._caption` when the main frame
gets a new document, so beats after a mid-take `goto()` log `""` instead of a
line that is not on screen. The artifact stopped lying.

## What is still worth saying out loud

The author does not learn that it happened. A storyboard that captions, then
navigates, then holds for eight seconds now records eight seconds of beats with
no caption at all — which is *true*, and is also very likely not what the
author meant. Before the fix the timeline hid the mistake behind a stale line;
after it, the timeline shows an empty column and still does not name the cause.

The cheap version is a `take issue` (not fatal) recorded at the moment the
caption is dropped, saying which line was lost and to which URL, so
`timeline.md`'s Issues section carries it and `strict=True` is unaffected.

## Measurement

On `examples/ticket-queue`, a storyboard of the shape `caption(...)`,
`goto("/x")`, `pause(3)` records 2 beats with `caption: ""` and no issue of any
kind. The recorder knows the line it dropped and the document it dropped it
for at the moment it drops it; neither reaches any artifact.

## Not included

Restoring the bar after the load. Re-rendering a caption the author did not ask
for again puts words on screen nobody wrote at that moment, which is the
failure mode this whole area exists to avoid.

Issue #134, opened 2026-07-28
A mid-take goto() leaves self._caption stale, so every later beat logs a caption that is not on screen


rogvid:
Found by an architecture review of `demo_recording`, then verified by reading
the code. Unfiled until now.

## The defect

`self._caption` is initialised at `core.py:2722` and assigned in exactly one
place — `caption()`, at `core.py:4782`. `_beat()` stamps it into every beat that
does not pass an explicit caption:

```python
"caption": self.scrub(self._caption if caption is None else caption),
```
— `core.py:3614`

The caption bar is a DOM element: `_CAPTION_JS` creates `#__demo_caption` in the
document (`core.py:49-52`). A full page load destroys it. `goto()`
(`web.py:1821`) performs exactly that navigation and **never clears
`self._caption`**.

So after a mid-take `goto()`, the bar is gone from the screen and the Python
attribute still holds the old line. Every subsequent beat's `caption` field
reports text that is not on screen.

## Why it matters more than it looks

The field is *defined* as the thing it stops being:

- `SKILL.md:695` — "`caption` is the line on screen during the beat"
- `core.py:785` — "the caption text on screen during the beat"

And it propagates. The same string reaches issue records (`core.py:3403`) and
every evidence document's `beat.caption` (`core.py:3925-3929`). `timeline.json`
and `timeline.md` are **committed to git** (`SKILL.md:658`), so the false
caption outlives the take that produced it and is what a reviewing agent reads
when the video is gone.

This is #91's shape — every artifact agreeing and all of them wrong — reachable
from an ordinary, documented storyboard rather than from an exotic one.

## It is currently an author obligation, and that is the bug

`SKILL.md:316` already says the caption "dies on full page loads, survives SPA
routing — clear before navigating either way." That instruction is correct and
unenforced. An author who forgets gets no warning, no issue recorded, and a
committed artifact that misreports. `core.py:1363-1365` cites the same fact as
the reason review frames ship uncaptioned — so the package knows the bar dies
and only acts on it in the frames path.

## Fix

Clear `self._caption` where the navigation is already noticed. `_note_navigation`
(`web.py:1164`) is on both `framenavigated` and `frameattached` for the paint
gate; the same signal is what invalidates the caption. Setting it to `""` there
covers `goto()`, link clicks, `go_back()`, form submits, `location.href` and
meta refresh in one place, rather than patching `goto()` alone.

SPA routing keeps the bar and must keep the string, so the clear belongs on the
document-replacing case only.

[...]

PR #216, opened 2026-08-01
Grade the caption clear against a real browser, and say what it dropped (#179, #180)

### LINKED ISSUE #179: The caption clear on a document load has no real-browser arm — the event name is taken on trust
Follow-up to #134, filed by the PR that fixed it.

## What is graded today

`tests/unit`'s `CaptionTruth` builds a `Recorder`, hands
`Recorder._watch_page` a page object that records what was subscribed to it,
fires `domcontentloaded` at the handler Playwright would call, and asserts the
next beat's `caption` is `""` and that the pre-navigation line does not appear
anywhere in the beats recorded after it. Four graded assertions, four
registered injections.

## What is not

**That a real Chromium fires `domcontentloaded` where this claims it does.**
The unit arm fires the event itself, so it proves the handler and the
subscription, and takes Playwright's contract on trust for the rest. The
reasoning behind picking that event over `framenavigated` is in
`web.py:_watch_page` and comes from reading `playwright-core`'s
`frameDispatcher` — `framenavigated` is dispatched for same-document history
navigation too, which is why clearing there would take a caption off the beat
log while it is still on screen.

Unverified against a browser, therefore:

- that a `goto()` mid-take delivers `domcontentloaded` before the *next* beat
  opens (the unit arm cannot observe ordering against a real navigation);
- that a link click, `go_back()`, a form submit, `location.href` and a meta
  refresh all deliver it, which is the claim that made the clear belong on the
  navigation signal rather than inside `goto()`;
- that an SPA route change does **not** deliver it.

## Acceptance

A `tests/smoke` arm that captions, `goto()`s, records further beats, and fails
if the pre-navigation line reappears in `timeline.json` — plus the SPA half,
which needs a fixture route that pushes history without loading a document.
Seen to fail with the clear reverted.

## Not included

Anything about the caption *bar* being restored after a load. It is not, by
design; this is only about the log agreeing with the screen.

rogvid:
Closes #179 and #180 — the two halves of what the beat log says about a caption
when the document is replaced.

## #179 — the event name is no longer taken on trust

**The claim in doubt was about *when Chromium sends* `domcontentloaded`, and
`tests/unit` fires it itself.** That is the catalogue's *check that shares the
bug's blind spot* exactly: it grades the handler and the subscription and goes
green under either event name. So the answer was measured instead of reasoned
about.

A probe (not committed — a re-measurement lands in the table) served a fixture
with a link, a GET form, a `location.href` button, a `history.pushState`
button, an iframe and a `<meta http-equiv=refresh>`, put the shipped
`Recorder._watch_page` subscription on a real page, and recorded the ordered
page events each way of leaving produced.

**Byte-identical on three builds — Chromium 136.0.7103.25 (Playwright 1.52),
147.0.7727.15 (1.59) and 151.0.7922.34 (1.62)** — the same three #202 is about,
where one build disagreeing is what the exercise exists to notice.

| what the storyboard did | page events, in order | caption |
|---|---|---|
| `goto("/b.html")` | `framenavigated`(main), `domcontentloaded`, `load` | cleared |
| click an `<a href>` | `framenavigated`(main), `domcontentloaded`, `load` | cleared |
| `go_back()` across documents | `framenavigated`(main), `framenavigated`(child), `domcontentloaded`, `load` | cleared |
| submit a GET form | `framenavigated`(main), `domcontentloaded`, `load` | cleared |
| assign `location.href` | `framenavigated`(main), `domcontentloaded`, `load` | cleared |
| `<meta http-equiv=refresh>` | two documents, **one `domcontentloaded` each** | cleared |
| `reload()` | `framenavigated`(main), `framenavigated`(child), `domcontentloaded`, `load` | cleared |
| `history.pushState` | `framenavigated`(main) **only** | kept |
| `go_back()` over that pushState | `framenavigated`(main) **only** | kept |
| an iframe loading a document | `framenavigated`(**child**) **only** | kept |

Every claim #179 listed, answered: fired once per new document (the meta
refresh is the only case that can show it — two documents, two events); only
for the main frame (the iframe produces a child-frame `framenavigated` and no
page-level document event at all); delivered for a link click, `go_back()`, a
form submit, `location.href` and a meta refresh; and **not** delivered for an
SPA route change.

Those streams are now `NAVIGATIONS` in `tests/unit`, replayed through the
production subscription and read off the **beat log** — never `rec._caption`,
which is the field #134 was about.

## #180 — the line does not go quietly

`Recorder._note_document_replaced` now records a `caption_lost` issue naming
the line and the document that replaced it, so `timeline.md`'s Issues section
carries it. New published kind in `ISSUE_KINDS`; deliberately **not** in
`STRICT_KINDS` — a storyboard mistake is not the app saying it is broken.

Confirmed against a real Chromium 151 with the fix in place, same probe:

```
goto           | caption '' | issues [('caption_lost', 'http://127.0.0.1:40263/b.html', 'The dashboard shows every open order.')]
link_click     | caption '' | issues [('caption_lost', 'http://127.0.0.1:40263/b.html', 'The dashboard shows every open order.')]
form_submit    | caption '' | issues [('caption_lost', 'http://127.0.0.1:40263/b.html?', 'The dashboard shows every open order.')]
meta_refresh   | caption '' | issues [('caption_lost', 'http://127.0.0.1:40263/refresh.html', 'The dashboard shows every open order.')]
spa_push_state | caption 'The dashbo…' | issues []
iframe_load    | caption 'The dashbo…' | issues []
```

No deadlock risk added: the handler reads `page.url`, which is the URL the
connection already told this process about and costs no round trip.

## Every assertion, seen to fail

`./tests/unit --fault-inject` — **all 103 injections caught**, 15 of them new or
re-pointed. The interesting ones:

**The injection #179 is about** — hang the clear on `framenavigated` instead.
Green before this PR, red now, and red *because a real pushState route change
sends that event and no document event at all*:

```
FAIL: test_every_measured_same_document_navigation_keeps_the_line (navigation='spa_push_state')
AssertionError: Lists differ: [''] != ['The dashboard shows every open order.']
- ['']
+ ['The dashboard shows every open order.'] : history.pushState to a new route,
  no document loaded left the caption bar in the document, and the beat log
  says it went away

FAIL: … (navigation='spa_back')
… : page.go_back() over that pushState — same document left the caption bar in
  the document, and the beat log says it went away

FAIL: … (navigation='iframe_load')
… : an iframe loading a document while the page stands still left the caption
  bar in the document, and the beat log says it went away
```

**#180's own** — `return` before the `_note_issue`, i.e. the clear without the
record, which is exactly what shipped before:

```
FAIL: CaptionLost.test_the_line_a_load_took_off_the_screen_is_recorded_as_a_problem
AssertionError: 0 != 1 : recorded []

FAIL: CaptionLost.test_the_lost_line_reaches_the_document_a_reviewer_reads
AssertionError: Lists differ: [] != ['caption_lost']

FAIL: CaptionLost.test_a_second_load_does_not_report_the_line_a_second_time
AssertionError: 0 != 1 : recorded []
```

**The other direction** — one issue per load rather than per lost line, which
would bury the real one under a storyboard that navigates six times (every take
opens with a document load):

```
PASS  break: every document load records a lost caption, including the ones with none
      FAIL: CaptionLost.test_a_load_with_no_caption_up_records_nothing
      FAIL: CaptionLost.test_a_second_load_does_not_report_the_line_a_second_time
```

**The table graded from its own side**, because a sweep over data is only as
good as the data:

```
PASS  break: a same-document route change is filed as a document replacement
      FAIL: MeasuredNavigations.test_every_measured_document_load_takes_the_line_off_the_beat_log (navigation='spa_push_state')
      FAIL: MeasuredNavigations.test_the_measured_table_covers_every_way_issue_179_names

PASS  break: the measured table stops covering the iframe #179 names
      FAIL: MeasuredNavigations.test_the_measured_table_covers_every_way_issue_179_names
```

**The exception path**, which is where a guard usually is not tested — the URL
is read off a page in the middle of replacing itself:

```
PASS  break: reading the new document's URL is allowed to throw out of the callback
      ERROR: CaptionLost.test_a_page_too_far_gone_to_name_still_reports_the_line
```

**The rest**: the clear removed; the subscription removed; the message dropping
the line; the message dropping the URL; `caption_lost` out of `ISSUE_KINDS`;
`caption_lost` moved *into* `STRICT_KINDS`.

### Two claims removed because the injections refused them

Both were written into the manifest and both proved unsupported — recorded here
rather than quietly deleted:

- `nothing subscribes to the document being replaced` does **not** fail
  `test_every_measured_same_document_navigation_keeps_the_line`.
  `framenavigated` is still subscribed, so that sweep's own control still holds
  and the caption still survives.
- The `framenavigated` swap does **not** fail
  `test_every_measured_document_load_takes_the_line_off_the_beat_log`. Every
  stream that replaces a document carries a `framenavigated` too, so the caption
  still clears there. **Only the same-document half can tell the two names
  apart** — which is the whole reason it had to be measured.

## Stated limits

- **The suite records three browsers; it does not drive one.** A build that
  started sending a document event for a `pushState` route change, or none for a
  form submit, would leave `tests/unit` green. Written into the file's *What this
  suite does NOT cover*.
- **`tests/smoke` has no arm for this.** That file is owned by another agent
  this session, so the arm #179's acceptance asks for is described in a comment
  below rather than written here.
- **A click-driven navigation is one beat late.** Measured, all three builds:
  `goto()` waits for the load, so the caption is gone when it returns, but a link
  click, a form submit and a `location.href` return after `framenavigated` alone.
  A beat stamps its caption when it *opens*, so the first beat after a
  click-driven navigation still carries the line.
  `MeasuredNavigations.test_a_load_the_verb_did_not_wait_for_reaches_the_log_one_beat_late`
  **pins that as it is today** (`[STALE, ""]`), so closing it has to be a
  deliberate edit here. Filed with its measurement as #214.
- Not touched, by #180's own exclusion: restoring the bar after a load.

## Verification

`./tests/unit` (164) · `--fault-inject` (103/103) · `--budget` (599/600,
unchanged) · `./tests/ci-unit` (49) · `./tests/lint` · `mypy 1.13.0` — green.
`./tests/smoke --web-only` — see the comment below.

## Scope note

`timeline.py` is touched beyond the stated scope of `core.py` / `web.py` /
`tests/unit`, minimally and unavoidably: a kind outside `ISSUE_KINDS` is refused
by smoke's `check_issues`, so a new one has to be published there. The change is
one tuple entry plus the three prose lines that enumerate the kinds.

### DIFF
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -247,7 +247,7 @@ exception, or a non-zero exit. Both are
 |---|---|
 | `goto(path)` | Navigate (relative to base_url); waits for networkidle, but gives up after 10 s for apps that poll |
 | `pause(s)` / `shot(name)` | Hold the frame / capture `images/<name>.png` |
-| `caption(text)` | Narrator line at the bottom; `""` clears; dies on full page loads and the beat log clears with it, survives SPA routing — re-caption after a load |
+| `caption(text)` | Narrator line at the bottom; `""` clears; dies on full page loads — the beat log clears with it and records a `caption_lost` issue naming the line — survives SPA routing; re-caption after a load |
 | `caption(text, ac="AC-3")` / `shot(name, ac="AC-3")` | Tag this beat with the acceptance criterion it is there to demonstrate. Needs `Recorder(criteria={...})`; a tag naming an undeclared criterion is refused. See [reference/review.md](reference/review.md). |
 | `hold(min_s=1.5)` | Keep the current frame up until the current caption's narration finishes (min `min_s`). Use after a spotlight/action so the emphasis rides the whole spoken line instead of flashing. See **Pacing and perception** below. |
 | `move_to` / `click` / `click_fast` / `scroll_to` | Visible cursor motion; `click_fast` for elements that re-render continuously |
--- a/skills/demo-video/helpers/demo_recording/timeline.py
+++ b/skills/demo-video/helpers/demo_recording/timeline.py
@@ -256,7 +256,7 @@ def _cap_text(text: str, limit: int) -> tuple[str, int]:
 #   message  str   — one human-readable line
 #   plus kind-specific keys: `url`/`line` (console), `url`/`method`
 #   (request_failed), `url`/`status` (http_error), `exit_code`/`command`
-#   (nonzero_exit)
+#   (nonzero_exit), `lost_caption`/`url` (caption_lost)
 #
 # `t` is when the problem was *observed*, which is not always when it happened:
 # Playwright's sync API only delivers page events while it is being called, and
@@ -280,13 +280,15 @@ ISSUE_KINDS = (
     "request_failed",   # a request that never got a response at all
     "http_error",       # a response with status >= 400
     "nonzero_exit",     # a TerminalRecorder run() whose command failed
+    "caption_lost",     # a page load took the caption bar off the screen
 )
 
 # What `strict=True` refuses to pass: the app saying, in its own voice, that it
-# is broken. `console_warning`, `request_failed` and `http_error` are recorded
-# but not fatal on their own — a warning is not a failure, and a request the
-# storyboard never depended on is the recorder's business to report, not to
-# veto.
+# is broken. `console_warning`, `request_failed`, `http_error` and
+# `caption_lost` are recorded but not fatal on their own — a warning is not a
+# failure, a request the storyboard never depended on is the recorder's
+# business to report rather than to veto, and a caption dropped by a
+# navigation is the storyboard's mistake and not the app's.
 #
 # In practice that distinction is narrower than it looks, and deliberately so:
 # Chromium writes its own "Failed to load resource: …" line to the console for
@@ -554,9 +556,9 @@ def render_timeline_md(doc: dict) -> str:
             "## Issues",
             "",
             f"{total} recorded while this take ran — console errors, failed "
-            f"requests, and non-zero exit codes, each attributed to the beat "
-            f"it fired during. A demo can look perfect and still be a "
-            f"recording of a broken app.",
+            f"requests, non-zero exit codes, and captions a page load took "
+            f"off the screen, each attributed to the beat it fired during. A "
+            f"demo can look perfect and still be a recording of a broken app.",
             "",
         ]
         # A bullet list rather than a table on purpose: a table row starting
--- a/skills/demo-video/helpers/demo_recording/web.py
+++ b/skills/demo-video/helpers/demo_recording/web.py
@@ -527,9 +527,12 @@ class Recorder(_DemoBase):
         and empty of issues. The base now seals `_watch_page` and calls this,
         so the same mistake is a `TypeError` while the module imports.
 
-        Both callbacks below only set an attribute. Playwright delivers page
-        events on the same thread that is blocked in a Playwright call, so
-        calling back into the API from one of them is a way to deadlock a take.
+        Neither callback below calls into Playwright. Page events are
+        delivered on the same thread that is blocked inside a Playwright call,
+        so calling back into the API from one of them is a way to deadlock a
+        take — `_note_document_replaced` writes an issue and reads `page.url`,
+        which is the last URL the connection already told this process about
+        and costs no round trip.
         """
         # **Kept deliberately when the masking went** — #142's carve-out.
         # Written for the paint gate, which is gone; nothing reads the flag.
@@ -545,6 +548,14 @@ class Recorder(_DemoBase):
         # clear a caption that is still on screen — the same lie in the other
         # direction. `domcontentloaded` is fired once per new document, and
         # only for the main frame, so an iframe loading does not touch it.
+        #
+        # That last sentence was read off `playwright-core` and is now
+        # measured: `NAVIGATIONS` in `tests/unit` holds the page-event stream
+        # a real Chromium delivers for `goto`, a link click, `go_back`, a form
+        # submit, `location.href`, a meta refresh, a reload, two same-document
+        # SPA route changes and an iframe load, identical on Chromium 136, 147
+        # and 151, and the suite replays each one through this subscription
+        # (issue #179).
         page.on("domcontentloaded", self._note_document_replaced)
 
     def _start(self) -> None:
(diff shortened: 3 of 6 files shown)


rogvid:
## The `tests/smoke` arm #179's acceptance asks for — described, not written

`tests/smoke` is owned by another agent this session, so this is the arm to add
rather than a diff. It is the half `tests/unit` structurally cannot do: driving
a real browser rather than replaying a recording of one.

**Where.** The existing `web` arm's fixture, plus two routes:

- `/nav-b.html` — any second document with a distinguishable `<h1>`.
- a same-document route change on the *existing* fixture page: a button whose
  handler is `history.pushState({}, '', '/route-2')` and nothing else. No
  document loads, so this is the half the whole event-name decision rests on.

**Storyboard.**

```python
rec.goto("/")
rec.caption("The dashboard shows every open order.")   # the line under test
rec.pause(0.5)
rec.click("#to-nav-b")        # a real <a href>, so the load is the browser's
rec.pause(1.5)                # beats recorded after the document was replaced
rec.caption("A second line, after the load.")
rec.pause(0.5)
rec.click("#push-route")      # pushState — the bar is still in the document
rec.pause(1.5)
```

**Assertions, all read off `timeline.json` and none off the recorder:**

1. No beat with `t_start` after the click that navigates carries the
   pre-navigation line — searched over `json.dumps(beat)`, not `beat["caption"]`,
   so a stale copy under any key fails.
2. **The control that stops it being vacuous:** at least one beat *before* that
   click carries the line. Without it, a recorder that stamped `""` on every
   beat passes.
3. Every beat after the `pushState` click carries `"A second line, after the
   load."` — the direction that fails if the clear is moved to `framenavigated`,
   and the only one that can tell the two event names apart.
4. Exactly one `caption_lost` issue, its `lost_caption` equal to the first line
   and its `url` ending `/nav-b.html` (#180).
5. `timeline.md`'s `## Issues` section contains that line.

**Seen to fail with the clear reverted** — revert `lost, self._caption =
self._caption, ""` to `lost = self._caption` and (1) goes red; move
`page.on("domcontentloaded", …)` to `framenavigated` and (3) goes red.

**One measured caveat for whoever writes it:** put a `pause()` between the
navigating click and the assertion window, or expect the *first* beat after the
click to still carry the line. `goto()` waits for the load; a click returns
after `framenavigated` alone and the document event lands during the next
Playwright call. That is #214, and this arm should not be written in a way that
either hides it or is broken by fixing it.

---

## Documentation to apply (files outside this PR's scope)

**`skills/demo-video/SKILL.md` line ~250** — the `caption(text)` row. Today's
text is not false, but it now under-reports: the beat log does more than clear.
Same line count, so the 599/600 budget is unchanged.

Replace:

```
| `caption(text)` | Narrator line at the bottom; `""` clears; dies on full page loads and the beat log clears with it, survives SPA routing — re-caption after a load |
```

with:

```
| `caption(text)` | Narrator line at the bottom; `""` clears; dies on full page loads — the beat log clears with it and records a `caption_lost` issue naming the line — survives SPA routing; re-caption after a load |
```

**`skills/demo-video/reference/failures.md`** — the `kind` table. Add a row
after `nonzero_exit`:

```
| `caption_lost` | a page load took the caption bar off the screen; the beat log clears with it and the issue names the line and the new URL | no |
```

**`tests/README.md`** — if that file's *Known gaps* section is being maintained,
#214 belongs in it: *a click-driven navigation clears the caption one beat later
than a `goto()` does, because a beat stamps its caption when it opens and the
document event has not been delivered yet; measured identical on Chromium 136,
147 and 151; pinned by
`MeasuredNavigations.test_a_load_the_verb_did_not_wait_for_reaches_the_log_one_beat_late`.*

rogvid:
## `tests/smoke` — what was run, and what it says

**CI's `smoke (record against the fixture app)` job passed on this branch.** That
job runs the *whole* suite (`tests/smoke --out-dir …`, every arm, not
`--web-only`) on a clean runner. All seven checks green:

```
typecheck (mypy): success
workflows (actionlint): success
lint (ruff, shellcheck): success
unit (demo-video CI helpers): success
smoke (record against the fixture app): success
secrets (gitleaks): success
unit (demo-video recorder): success
```

That is the regression answer for this change, and it also covers the arm that
would have been most sensitive to it: `check_healthy` requires `issues == []` on
the strict narration take, so a `caption_lost` recorded where no caption was
lost would fail it.

**Locally, `./tests/smoke --web-only` is red on this box on both branches, with
different failures each time** — recorded here rather than glossed:

On this branch (twice, same two):

```
smoke: FAILED
  - web: the closing caption could not be timed against demo.mp4 — the caption
    arrives 0.40s into a 1.2s run-up, leaving no frames before its fade to use
    as a baseline — the video has slid further under the beat log than the
    search window allows for (see issue #18)
  - web: beat-23.png should show the caption 'Keys, not just clicks.' and its
    caption band is only 3.52 from the take's caption-off baseline —
    beat-27.png, which should show no caption at all, is 2.89 from it. The two
    are 0.64 apart, under the 6.0 floor
```

On `origin/main`, unmodified, same machine, minutes later — red too, and on a
*different* arm:

```
smoke: FAILED
  - overlay-cleared: after the documented `interlude("")` the app rect is still
    63.86 mean luma from where it was before the scrim went up — 100% of the
    63.93 the take that never cleared it keeps, over the 20% bar
```

Both are frame-timing measurements against a recording, and both cite open
issues about exactly that (#18's ~0.57 s of wall time the screencast drops,
#162's overlay). This machine is running several agents recording video at once;
the smoke lock serializes smoke suites, not the rest of the load. A red run
under contention is worse than no run — so the reading to trust here is CI's,
which is green, and `origin/main` failing a *different* arm on the same box is
the control that says so.

Nothing in this change touches frame timing, the overlay, or the encode: the
diff outside `tests/unit` is one attribute assignment split into two statements,
one `_note_issue` call in an event callback, one tuple entry and four prose
lines.

PR #234, opened 2026-08-01
Merge capture_clock into a stitched demo, per capture (#225)

rogvid:
Closes #225.

## Acceptance criterion

> A stitched `demo.mp4` recorded on a host whose clock stepped during one of
> its parts carries enough in its merged `timeline.json` for a reader to place
> any beat in the video to within the residual `tests/smoke` measures per
> segment, **without reading the parts**. Graded by an entry in
> `tests/smoke-inject` against `--segments-only`.

## What it does

`stitch()` merges `capture_clock` the way it merges `beats`: each part's steps
moved onto the stitched clock by that part's ffprobe offset. Two things make
the result usable, and both are about *which steps apply to which beats*:

- **every step names the `segment` that measured it, and that is the
  attribution to trust.** A step is sampled for as long as its capture runs,
  which is longer than the video that capture produced — losing wall time is
  the whole phenomenon — so a step's merged `t` can legitimately fall past the
  next part's boundary. This is not hypothetical: one of the runs below has a
  -876 ms step at 7.2 s of a part whose video is 6.88 s.
- **`boundaries`** records where each capture starts on the stitched clock.

The correction for a beat is its own segment's steps up to its `t_start` —
never `total`, and never an earlier part's, whose lost wall time is already in
the offset `stitch()` laid it out by. Applying it twice over-corrects by a
whole step.

The merged record is **null** unless every part carries a usable one; each
`segments` record also keeps its own part's record untouched, so a null at the
top does not take the parts that *were* measured with it.

`timeline.py`'s envelope documentation gains `capture_clock` (it was documented
only where `core.py` builds it), and `reference/limits.md` gains the
stitched-demo rule.

## What I measured

Verdict lines, in full:

| suite | verdict |
|---|---|
| `./tests/unit` | `Ran 187 tests in 0.485s` -> **OK** |
| `./tests/unit --fault-inject` | **All 113 injections were caught.** |
| `./tests/ci-unit` | `Ran 49 tests in 0.129s` -> **OK** |
| `./tests/lint` | **All checks passed!** / 14 files already formatted |
| `ruff format --check .` + `ruff check .` (repo root) | 14 files already formatted / All checks passed! |
| `./tests/unit --budget` | `SKILL.md: 600 lines (~14,114 tokens), budget 600 lines` — unchanged |
| `./tests/smoke --segments-only` | **smoke: PASSED** |
| `./tests/smoke-inject --self-test` | **Every guard refused what it exists to refuse.** |
[...]

Issue #226, opened 2026-08-01
Narration is mixed on the wall clock while pixels ride the screencast's, so a stepped host lags the voiceover


rogvid:
From [#215](https://github.com/rogvid/skills/issues/215), and the half of
[#18](https://github.com/rogvid/skills/issues/18) that a *viewer* notices
rather than a manifest reader.

`core._convert` mixes each narration clip at the offset recorded in `_lines`,
which is `time.monotonic() - _t0` — the beat log's clock. The video it is mixed
into is on the host's wall clock, because that is what Chromium stamps
screencast frames with. So every wall-clock step the host takes during a take
moves the pixels and leaves the audio where it was: a demo whose host stepped
-0.8 s has its voiceover 0.8 s late against the caption for the rest of the
take.

#18's second comment measured exactly this before the cause was known: **+0.70 s
of audio lag on all three lines** in a take that "stalled in the run-up",
against +0.11 to +0.14 s in a clean one. That 0.70 s is a wall-clock step.

The recorder now *knows* this happened — `capture_clock` in `timeline.json`
records each step with its offset and size — so unlike everything else about
#18 this one is correctable at the point it is produced, before the file is
written:

    adelay = line_offset + steps_before(line_offset)

with the steps taken from the same `_CaptureClock` the timeline already
carries. Nothing has to be estimated.

## Why it is filed rather than done

It changes the audio of every recording made on a stepping host, and the
correction wants its own measurement: a synthetic tone through the real
`adelay`/`amix` path, audio onsets and caption transitions read out of the same
mp4, on a take whose host clock is known to have stepped and one whose has not.
#18's comment describes that harness; it is not in the repo.

## Acceptance

On a take whose host wall clock stepped during recording, the audio onset of
each narration line sits the same distance from its caption's appearance as it
does on a take whose clock did not — measured off the encoded mp4, not
computed. **Deliberately out of scope:** #224 (the terminal arm's extra,
unattributed loss), which would move the pixels again and is not a narration
question.

rogvid:
**Premise confirmed, and the input this depends on is now trustworthy.** #250
measured that the video really is on the host's wall clock while the beat log is
on the monotonic one: six takes, 38 caption transitions read by luma off
`demo.mp4`, uncorrected up to **−1.50 s** from the log by 13.5 s into the take,
and **all 38 within 101 ms** once corrected by the wall-clock offset before each
beat. So `adelay = line_offset + steps_before(line_offset)` is the right rule.

[...]

Issue #224, opened 2026-08-01
The terminal take loses ~1s more than its host clock step explains, and the correction cannot see it


rogvid:
Split out of [#215](https://github.com/rogvid/skills/issues/215) while fixing it, and this is the part the fix does **not** cover.

## What #215 established

Chromium stamps every screencast frame with the host's *wall* clock
(`Page.screencastFrame.metadata.timestamp`, a `Network.TimeSinceEpoch`), and
Playwright's `FfmpegVideoRecorder` turns that timestamp straight into the
frame's position in the webm. The beat log is `time.monotonic()`. This box's
wall clock steps **-0.75 to -0.81 s every 32.2 s**, so a take that contains a
step loses exactly that much video and leaves the log where it was.

`tests/smoke` now measures the same clock beside every recorded take and
subtracts it before grading. On the web arm and on both segments of the
stitched demo the correction is exact: residual 20-140 ms, over ~20 runs,
including runs whose step landed between the two timing probes.

## What this issue is

The terminal arm is not exact. Four `--terminal-only` runs, with the
recorder's own `capture_clock` read back out of each take:

| run | measured step | first probe | closing probe | result |
|---|---|---|---|---|
| 1 | none | -180 ms | -140 ms | PASSED |
| 2 | -800 ms at 1.8s | ~-1.94 s raw, unmeasurable in the window | over the bar | FAILED |
| 3 | -839 ms at 8.9s | -100 ms | **-900 ms after correction** | FAILED |

Run 3 is the clean shape. One step, measured, and this harness and the
recorder agree on its size and its time inside the storyboard's own window
(`check_capture_clock` compares both directions over `[0, last t_end]` and
passed). Subtract it and there is still **0.9 s** — about one more step — of
wall time missing from the video between 8.9 s and 11.95 s.

The take's *length* only lost the one step: `duration` was 17.4 s against
18.16 s for a step-free run of the same storyboard, a difference of 0.76 s.
So the file did not simply come out shorter by two steps — something between
the jump and the closing caption is compressed while the total is not.

## The candidate worth eliminating first

What ffmpeg does with the cluster timestamps Playwright writes after Chromium's
clock goes backwards. `FfmpegVideoRecorder._writeFrame` computes
`frameNumber = (ts - firstFrameTs) * 25` and `_emitFrame` writes
`frameNumber * 40 ms` as the cluster timestamp — so after a backwards jump the
timestamps it writes **decrease**, and the encode is a CFR `-r 25` output from
[...]

Issue #215, opened 2026-08-01
The screencast still loses 0.8s of a web take with TICKER_JS running, and sometimes mid-take


rogvid:
Measured while grading [#209](https://github.com/rogvid/skills/issues/209).
`TICKER_JS` was added so that storyboard idle could not make Chromium's
screencast drop wall time, and `tests/smoke`'s own comment records the
before/after that justified it: 3/3 idle takes stalled, 3/3 ticking takes did
not. On this box today, ticking takes stall.

## The measurement

Four `tests/smoke --web-only` runs, one at a time behind the machine lock, on
`main` @ d54bcf3. The number is how far `demo.mp4` sits **ahead** of the beat
log at each of the two caption probes — read out of the arm's own output, and
where the measurement could not be made, computed from the reported arrival
(`ALIGN_PRE_S` minus the arrival offset).

| run | loadavg | result | first probe | closing probe |
|---|---|---|---|---|
| 1 | 1.25 | FAILED | -880 ms | -800 ms |
| 2 | 2.05 | PASSED | -80 ms | 0 ms |
| 3 | 2.85 | FAILED | -120 ms | -800 ms |
| 4 | 19.8 (16 busy workers) | FAILED | -880 ms | -840 ms |

Bimodal: either under 120 ms, or 800-880 ms. Nothing lands in between.

`start_ticker()` asserts the ticker is attached, animating, and spared by the
determinism rule on every one of these takes, and none of those assertions
fired. The ticker was alive and the capture still lost the time.

## Two shapes, and the second is the interesting one

**Uniform** (runs 1 and 4): both probes read the same, within 80 ms. This is
the recorder's own un-tickerable setup window lost whole — the shape
[#18](https://github.com/rogvid/skills/issues/18) describes and
`MAX_CAPTURE_LOSS_S` exists to tolerate. It is now measured at 0.80-0.88 s,
against the -600 ms that constant was calibrated from.

**Mid-take** (run 3): the first probe reads -120 ms and the closing one
-800 ms. 680 ms went missing *between* them, which the ticker is supposed to
make impossible. This is the reading that says the mitigation is not holding,
and it is also the one no widened bar can absorb: `MAX_SKEW_DRIFT_S` is 250 ms
and its whole argument is that a capture stall moves both probes together.

## Why this is filed rather than fixed in the suite

`tests/smoke` cannot fix this and should not paper over it. See
`tests/README.md`, **Known gaps**, for what was changed instead (the search
[...]

PR #247, opened 2026-08-08
Cut the review sheet's frames on the clock the video is actually on (#229)

### LINKED ISSUE #229: Nothing applies capture_clock: the review sheet's frames are cut on the uncorrected beat log, and timeline.md never mentions the field
Split out of [#225](https://github.com/rogvid/skills/issues/225), where the
merged `capture_clock` was added but nothing was made to *use* it.

The recorder measures the host's wall-clock steps and writes them into
`timeline.json` as `capture_clock`, and since #225 a stitched demo carries them
too, per capture. Two consumers inside this repo still ignore the field:

- **`write_beat_frames` cuts every review frame at the beat's midpoint on the
  raw beat log** (`frames.py`, `middle = (t_start + t_end) / 2`). On a host that
  stepped its clock during the take, that instant is off by the sum of the
  steps before it — 0.75–0.81 s per step on the WSL2 box #215 measured, which
  at 25 fps is ~20 frames. The review sheet is how a demo is reviewed in this
  skill (nobody watches the video), so a frame cut 20 frames early is the demo
  as far as a reviewer is concerned.
- **`render_timeline_md` never mentions `capture_clock` at all.** A human
  reading `timeline.md` sees beat timestamps with nothing saying the video slid
  under them. The recorder does warn — on stderr, at exit, minutes and several
  thousand lines before anyone opens the sheet.

## What it needs

`write_beat_frames` corrects the cut instant with the record: for a single take,
the steps up to that beat; for a stitched one, the steps of *that beat's own
segment* up to it (see the envelope docs in `timeline.py` — the segment is the
attribution, not the timestamp). And `timeline.md` says, when the total is
non-zero, that the clock stepped and by how much, next to the timestamps it
affects.

## Acceptance

On a take recorded while the host's clock stepped, the frame `beat-NN.png`
shows the beat's midpoint to within the residual `tests/smoke` measures — not
the step. Graded by an entry in `tests/smoke-inject`; `--segments-only`
already records a stitched demo and already writes the sheet.

**Deliberately out of scope:** #226 (narration), #224 (the terminal arm's extra
loss). Neither is a question about who applies the correction.

rogvid:
Fixes #229.

## Acceptance criterion

On a take recorded while the host's clock stepped, `beat-NN.png` shows the beat's midpoint to within the residual `tests/smoke` measures — not the step. Graded by a `tests/smoke-inject` entry against `--segments-only`.

## What changed

`write_beat_frames` cut every frame at its beat's midpoint on the raw beat log. The beat log is `time.monotonic()`; Chromium stamps each screencast frame with the host's **wall** clock. The recorder has measured the difference since #215 and written it as `capture_clock` — and until now **no line of the shipped package read it back**. `tests/smoke` corrected; the product did not. The harness knew what time it was and the artifact did not.

`capture_clock_shift` applies the record to a seek: for a single take, the steps up to that beat; for a stitched one, the steps of **that beat's own segment**, because a capture keeps sampling after its last frame and a merged `t` can legitimately fall past the next boundary. `timeline.md` now states when the clock stepped and by how much.

`tests/smoke` keeps its own independent implementation deliberately. Two implementations that agree are a check; one shared implementation is a restatement.

`reference/review.md` also said the beat log was wall-clock, and blamed idle screencast stretches — the explanation #215 falsified. Both corrected here rather than left standing beside the fix.

## The injection, and why it grades on any host

Deleting the correction is a **no-op wherever the wall clock never steps**, so that entry would pass against a recorder that ignores the record entirely — the catalogue's *environment-agreeing assertion*, and the trap #234 walked into and caught. Instead the entry makes the recorder report a step **nobody measured**, at `t=0` so it bears on every beat. A recorder that applies its record then cuts frames 500 ms from where this harness — which watched the real clock and saw nothing — puts them. A recorder that ignores its record cuts them where they were, and the entry goes unnoticed. Same grading on a stepping host and a steady one.

## Verification, and one honest gap

Browser-free, on this branch:

| suite | verdict |
|---|---|
| `./tests/unit` | `Ran 214 tests` -> **OK** |
| `./tests/unit --fault-inject` | **All 129 injections were caught.** |
| `./tests/lint` | **All checks passed!** / `10 python fence(s) parse` |
| `ruff format --check .` / `ruff check .` (repo root) | clean |
| `./tests/unit --budget` | `SKILL.md: 600 lines` — unchanged, net zero |
| `./tests/smoke-inject --list` | 43 entries, 12 arms, every entry still matches the tree |

**The `--segments-only` injection has not been seen to fail on this branch, and that is stated rather than glossed.** The authoring session was lost to an API limit before its evidence was recorded, and the attempt to regenerate it aborted — correctly:

```
ABORT: the clean --segments-only run failed, so nothing aimed at that arm can be
       graded — an assertion firing under injection would prove nothing if it
       fires without one.
```

The cause is the host, measured directly rather than guessed. Sampling `time.time()` against `time.monotonic()` for 31.6 s on this box right now:

```
steps over 50 ms: +10100, -10603, +10102, -10594, +10094, -10596,
                  +10095, -10587, +10086, -10599, +10099, -10595 ms
net wall-minus-monotonic drift: -2.999 s over 31.6 s
```

The wall clock is oscillating **+/-10 s every ~5.5 s**. That is a different animal from the 0.75-0.81 s steps #215 measured on the same box, and no take recorded here is currently gradeable. CI's clock is not this box's, so **`smoke-full` is labelled on this PR** and the `--segments-only` arm grades there.

Two things worth noting from that abort. The harness refused to grade rather than emitting a green — the `check_capture_clock` machinery from #227 doing exactly its job. And the design holds under an insult far past what it was built for: because `capture_clock` **measures** rather than assuming a bound, a host oscillating +/-10 s is still describable. A fixed tolerance would have been silently wrong.

## Provenance

The authoring agent was killed by an API session limit, and its work was later reset out of the working tree by another agent before it was ever committed. It was recovered from a saved diff, applied to a clean checkout of `origin/main`, and verified there — the suites above were run on the recovered tree, not inherited from the lost session.

## Out of scope

#224 (the terminal arm's extra unattributed loss) and #226 (narration mixed on the monotonic clock) are untouched, per the issue.

### DIFF
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -95,9 +95,9 @@ behind every one, are in [reference/limits.md](reference/limits.md).
 - **A frame is aimed at a beat, not stamped with one.** Chromium stamps frames
   with the host's **wall** clock, the beat log uses a monotonic one, so on a
   host whose clock steps an event lands earlier in `demo.mp4` — by the step,
-  once per step (0.75–0.81 s every 32.2 s on one WSL2 box). **No fixed bound
-  holds**; a longer demo accumulates more. `timeline.json`'s `capture_clock`
-  records each one: correct with it, or read a frame as *around* its beat.
+  once per step (0.75–0.81 s every 32.2 s on one WSL2 box). `frames/` is now
+  cut with `timeline.json`'s `capture_clock` and `timeline.md` says when the
+  clock stepped; a beat time *you* scrub to yourself is still uncorrected.
 - **Sixty seconds buys about twenty screens.** At the pacing floors below, a
   demo shows roughly one new screen every 3 s — 69% of a real 61.2 s take was
   a picture already shown. A feature spanning two surfaces does not fit the
--- a/skills/demo-video/helpers/demo_recording/frames.py
+++ b/skills/demo-video/helpers/demo_recording/frames.py
@@ -31,6 +31,16 @@ from .timeline import timeline_paths
 # it is before the verb has done anything. The midpoint is inside the beat by
 # construction, which is what `t_end` is on the record for.
 #
+# That midpoint is a *beat log* instant, and the seek is a *video* one. The two
+# are on different clocks — the log on `time.monotonic()`, the video on the
+# host's wall clock, which some hosts step by 0.75-0.81 s at a time (issues
+# #18, #215) — so the seek is the midpoint plus what `capture_clock` says the
+# clock had done by then. `capture_clock_shift` is that arithmetic and issue
+# #229 is why it is here: the recorder measured the steps, wrote them down,
+# and then cut every frame as though it had not. At 25 fps a 0.78 s step is
+# ~20 frames, and since nobody reviewing through this skill watches the video,
+# a frame cut 20 frames early *is* the demo to its reviewer.
+#
 # **These frames carry no caption, and that is deliberate.** The obvious next
 # step is to print each frame under the line that was on screen when its beat
 # ran, and it is not sound: the beat log and the video are on different clocks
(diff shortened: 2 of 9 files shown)


rogvid:
Rebased onto `75b4c7f` — #242 and #243 landed underneath this. One conflict, `tests/README.md`, in both places the manifest figures are quoted; #242 took the count 42 → 45 while this branch took it 42 → 43. Resolved by **running `--list` and quoting it** rather than adding the two deltas together, since `--self-test` verifies those figures against the real output and arithmetic would have been a guess.

```
46 entries, 12 arms, ~40.0 min of takes
every entry still matches the tree
16 of tests/smoke's 39 check function(s) are ungraded
```

Ungraded is **19 → 16** with #242's three, and this branch's entry is the 46th.

Verified on the rebased tree: `./tests/unit` → `Ran 214 tests` **OK**; `--fault-inject` → **All 129 injections were caught.**; `--self-test` → **Every guard refused what it exists to refuse.**; `./tests/lint` clean; `--budget` `SKILL.md: 600 lines`.

The `--segments-only` injection is still ungraded locally for the host-clock reason in the body above — that box is unchanged. `smoke-full` is labelled, so CI grades it.

rogvid:
Closing this. It should not have been opened, and the reason is worth recording rather than deleting.

This branch is a diff I recovered from the main worktree after the authoring agent was killed by an API limit. I checked for the work in commits, refs, stashes and worktrees, found it nowhere, and concluded the diff was the only copy. **It was not.** The agent was still alive; it reset the tree under me and then landed its actual conclusion as #245 and #246 — which is that this change is wrong and must not ship.

#245 measures it: on this host `capture_clock` is not a step the video followed, it is an **aliased sample** of a wall clock oscillating ±10 s every ~5.5 s, caught at 20 ms. And the video does not follow it — across eight `--segments-only` runs, a take with a recorded −1.7 s step puts its closing caption within 20 ms of the log, the same as a step-free one, when the documented consequence predicts 43 frames of slide. Applying the correction moved `beat-12.png` to 10.53 s instead of 12.27 s and it showed the screen before the caption arrived. **Three of four stepped runs went red on the one check that reads pixels rather than timestamps** — the artifact-lies shape, in the single document a reviewer of this skill actually reads.

So the acceptance criterion in #229 is not met by this patch, and CI going green here would have proved nothing: a runner's clock is steady, so the harmful case cannot arise there. That is the *environment-agreeing* trap one level up — not in an assertion, but in the decision to merge on a green tick.

What I got wrong, precisely: I verified the recovered work was **green**, and treated that as evidence it was **finished**. Green was never the question. The author had already run the measurement that green cannot see.

#246 carries the correct outcome — the measurement as a Known-gaps entry, no product change, and #245 with the workings. #229 stays open, and #245's checklist is what has to land before anyone tries again: corroborate a step against the encoded duration, sample faster than the excursion, and re-measure #215's 0.75–0.81 s, which is the same shape as an aliased oscillation with a different period and which the whole field rests on.

Issue #245, opened 2026-08-08
capture_clock is an aliased sample of an oscillating wall clock, so nothing can correct with it — and tests/smoke is already red about it


rogvid:
Found while building [#229](https://github.com/rogvid/skills/issues/229), which
asked `write_beat_frames` to correct a review frame's seek with `capture_clock`.
It cannot, because on this box today the field is not a measurement of a step
the video followed — it is an **aliased sample of a wall clock that oscillates
faster than the sampler**. The same is true of `tests/smoke`'s own `HostClock`,
which is why `main` is already red about it.

Everything below is measured on `main` @ `bae7f74`, Playwright 1.62.0,
Chromium 151.0.7922.34 — the same versions
[#215](https://github.com/rogvid/skills/issues/215) was measured on.

## 1. What the host's wall clock is actually doing

A standalone sampler — no browser, `time.time() - time.monotonic()` every
20 ms, the same rule `_CaptureClock` uses — for 180 s. **66 events, in 33
identical pairs:**

```
66 event(s) in 180s   (t, delta_offset, d_wall, d_monotonic)
   (2.577,  10.3475,  10.3678, 0.0204)
   (2.638, -10.5914, -10.5711, 0.0203)
   (7.823,  10.3572,  10.3780, 0.0209)
   (7.885, -10.5939, -10.5736, 0.0203)
   ...
   (177.890,  10.0847,  10.1054, 0.0207)
   (177.954, -10.5950, -10.5743, 0.0207)
```

`time.time()` jumps **+10.08 to +10.36 s and returns −10.58 to −10.60 s about
60 ms later, every ~5.5 s**, all night. `d_monotonic` is a steady 20 ms
throughout, so this is the wall clock and not the sampler being descheduled.

`_CaptureClock` samples at 20 ms while competing with Playwright for the GIL,
so it catches a *fragment* of that ~60 ms excursion. That is why a take records
a single step of **−1.72 to −1.81 s**, or **+10.36 s**, and never the ±10.1/10.6
edges the idle sampler sees. The number in `capture_clock` is an aliasing
artifact of the sampling rate, not the size of anything.

## 2. The video does not follow it

Eight `--segments-only` runs, four with a recorded step. The closing caption's
onset read straight off `demo.mp4` (caption-band luma, 40 ms resolution)
against where `timeline.json` puts that beat:

| run | recorded step | log | video | video − log | demo.mp4 |
|---|---|---|---|---|---|
[...]

Issue #244, opened 2026-08-08
limits.md says "Two limits live elsewhere" but determinism.md carries three


rogvid:
`skills/demo-video/reference/limits.md:18-21` tells the reader which boundaries are deliberately not in that file:

> Two limits live elsewhere and are not repeated. The recorder hides nothing that
> reaches the screen — the top of [SKILL.md](../SKILL.md) […]. A frozen clock
> changes what an app does and usually does it silently —
> [determinism.md](determinism.md).

Measured at `bae7f74`, before #243: `limits.md` contains **zero** occurrences of
`cursor` or `pointer`, while `determinism.md:126-135` already carried two cursor
boundaries the sentence does not name — that the dot is drawn only once the
pointer is somewhere it has not been (#186), and that a pointer verb landing on
exactly `(0, 0)` therefore draws no cursor. #243 adds a third beside them (a
pointer move into a document that has not loaded can be dropped outright, #230),
which is the right home for it — the cursor's limits should stay in one file —
but it widens the gap between the count and the contents.

So an author who reads `limits.md` cover to cover, as the file invites, comes
away believing the only thing living in `determinism.md` is the frozen clock,
and never learns that every boundary about the recorder's own cursor is over
there.

## Acceptance

`limits.md`'s pointer-elsewhere paragraph names the cursor boundaries as living
in `determinism.md`, in one clause, without restating them — a reader looking
for why their take has no cursor is sent to the right file from the file they
opened.

## Deliberately not in scope

Moving the cursor boundaries into `limits.md`, or duplicating them. They belong
in one place and `determinism.md` is that place; this is about the signpost, not
the content. Also out of scope: whether `SKILL.md` should link `determinism.md`
from its `Layout shifts strand the cursor` bullet (`SKILL.md:553-555`) — that
costs a line against a budget that is exactly full.

Issue #218, opened 2026-08-01
Nothing reaps the temp directories a failed smoke run keeps on purpose


rogvid:
Split out of [#105](https://github.com/rogvid/skills/issues/105), whose second
comment measured two sources of `/tmp/demo-video-smoke-*` accumulation and
whose body — and acceptance — only covered the first. PR #217 closes that half:
a run the machine lock refuses no longer creates a directory at all, and is
graded on it by `tests/smoke --lock-only`.

This is the other half, unchanged by that PR.

## What is left

**A failed run keeps its directory deliberately.** That is correct behaviour
and it is what makes a failure debuggable — `main()` only removes an ephemeral
`out_root` when `failures` is empty. But nothing ever reaps them, so every red
run in the project's history is still on disk.

Measured on this box just now, after #217's fix landed on a branch:

```
$ ls -d /tmp/demo-video-smoke-* | wc -l
134
$ du -sh /tmp/demo-video-smoke-* | tail -1
1.4M    /tmp/demo-video-smoke-zylrler3
```

The 1.4 MB directories are the failed takes; #105's comment measured 85
directories and 80 MB over three days of ordinary work, and it has kept
growing.

## Acceptance

A policy, whichever is cheaper to defend: keep-last-N, or prune on startup
anything older than a few days. What matters is that it is a *decision* rather
than the current implicit "never", and that the directory a run just failed in
survives it — the point of keeping them is the run somebody is about to look
at.

## Deliberately not in scope

The refusal path, which #217 fixed and grades.

## One thing to know before writing the reaper

`tests/smoke-inject`'s first `--lock-only` entry reproduces #105's bug on
purpose: it stages a copy of `tests/smoke` with the directory built before the
lock, and the refused child therefore creates one empty
`/tmp/demo-video-smoke-*` per injected run. That is inherent to reproducing the
bug, and it is a small, steady contributor to this pile — a reaper that removes
*empty* ones would clear it, or `check_lock_refusal()` could remove the
directories it has just proved should not exist.

PR #249, opened 2026-08-08
Reap the old smoke output directories, and point limits.md at the cursor (#218, #244)

rogvid:
Two small things, folded into one change rather than carried as issues.

## #218 — the reaper

`open_output` keeps the **five most recent** `/tmp/demo-video-smoke-*` directories and removes the rest, before it creates this run's own so that one is never a candidate.

**Two corrections to the issue's framing, found while implementing it.**

The pile is bounded by the host clearing `/tmp` on restart. #218 measured 134 directories; there are **zero** on this box today. So this is hygiene, not the leak the issue describes — worth doing, not worth what the issue implies.

And the issue does not mention the guard the reaper actually needs: `--allow-concurrent` means a directory minutes old may still be receiving a take. Nothing under an hour is touched, **and young directories spend the budget they occupy** — otherwise the floor rises with every concurrent run and the pile grows forever.

## The injections

The selection is a pure function, so the budget arithmetic is graded without staging real directories with faked mtimes — which would have graded the staging. Five cases, three injections, all seen to fail:

```
PASS  break: the reaper treats a directory a live run is writing to as reapable
PASS  break: young directories are kept without spending the budget they occupy
PASS  break: the reaper keeps the oldest directories instead of the newest
All 127 injections were caught.
```

The last two matter most: **both leave exactly the right number of directories on disk**, so a check that counted survivors would pass while the reaper deleted the one somebody was about to read.

One injection was rejected by the harness before it landed, and that is worth recording. I first broke the guard as `young = []`, which reads like removing the concurrency protection but does not — the age filter on `old` still excludes young directories, so nothing young could be reaped and only one of the two tests I had named failed. The exactly-once contract caught the mismatch. The guard actually lives in the `old` comprehension, and the injection now targets it.

## #244 — the pointer

`limits.md` said "Two limits live elsewhere" and named the frozen clock, while `determinism.md` carried three cursor boundaries it did not mention — four since #243. Named in one clause, not restated, so the cursor's limits stay in one file.

## Verification

| suite | verdict |
|---|---|
| `./tests/unit` | `Ran 198 tests` → **OK** |
| `./tests/unit --fault-inject` | **All 127 injections were caught.** |
| `./tests/lint` | **All checks passed!** |
| `ruff format --check .` / `ruff check .` (root) | clean |
| `./tests/smoke-inject --self-test` | **Every guard refused what it exists to refuse.** |
| `./tests/unit --budget` | `SKILL.md: 600 lines` — untouched |

No `tests/smoke` arm run locally: this host's wall clock is currently oscillating ±10 s every ~5.5 s (#245), so no recorded take is gradeable here. The reaper's behaviour is graded browser-free by the cases above; CI records the arms.

### DIFF
--- a/skills/demo-video/reference/limits.md
+++ b/skills/demo-video/reference/limits.md
@@ -14,10 +14,12 @@ newly found gap outside it is a line in this file, and an author who knows the
 boundary can work around it. Each section names the issue it supersedes, and
 carries that issue's numbers, so the evidence outlives the issue.
 
-Two limits live elsewhere and are not repeated. The recorder hides nothing that
+Some limits live elsewhere and are not repeated. The recorder hides nothing that
 reaches the screen — the top of [SKILL.md](../SKILL.md), and the reason there
 is no masking verb. A frozen clock changes what an app does and usually does it
-silently — [determinism.md](determinism.md).
+silently, and every boundary about the recorder's own cursor sits beside it —
+when the dot is drawn at all, why a verb landing on exactly `(0, 0)` draws
+nothing, and when a move can be dropped outright — [determinism.md](determinism.md).
 
 ## What the recorder will not notice about your app
 
--- a/tests/smoke
+++ b/tests/smoke
@@ -11025,6 +11025,60 @@ def main() -> int:
     return 0
 
 
+# How many previous ephemeral output directories survive a new run, and the
+# age under which one is never touched however many there are (issue #218).
+#
+# A failed run keeps its directory on purpose — that is what makes a failure
+# debuggable, and `main()` only removes an ephemeral `out_root` when there were
+# no failures. Nothing ever reaped them, so the pile was 134 directories when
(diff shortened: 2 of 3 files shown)


rogvid:
Rebased onto `8f21426` (#250). No conflict, and I checked that rather than trusting it — #250 rewrote most of `limits.md`, including the section this touches, but it left the pointer paragraph alone, so the two edits compose. The paragraph now reads correctly beside #250's rewritten timing section.

Rebased totals, against `origin/main` at 207 tests / 132 injections: **212 tests / 135 injections** — the +5 and +3 this change should cost, nothing lost to a merge that was only clean about text.

`./tests/unit` **OK**; `--fault-inject` **All 135 injections were caught.**; `./tests/lint` **All checks passed!**; `smoke-inject --self-test` **Every guard refused what it exists to refuse.**

Issue #240, opened 2026-08-01
--polish-only reaches three ungraded checks at 26 s and carries no entry at all


rogvid:
One slice of the mapping derived in the sibling issue for `--narration-only`:
of the 19 check functions `./tests/smoke-inject --list` prints as ungraded,
only four are expensive. These three are not, and one of them is currently
the only check function in the file with **no exercise anywhere but a
per-push take**.

## The slice

| check function | phases that reach it | cheapest arm | s/entry |
|---|---|---|---|
| `check_spotlight_transitions` | `run_spotlight` | `--polish-only` | 26 |
| `check_opening_card` | `run_terminal_opening` | `--polish-only` | 26 |
| `_check_video` | `run_spotlight`, `run_terminal_opening`, and `check_take` from `run_web`/`run_segments`/`run_terminal` | `--polish-only` | 26 |

`--polish-only` has **no entry at all** today, so this slice pays one clean
baseline (26 s) on top of 26 s an entry.

`tests/README.md` already states, in *Why the four middle arms stayed on the
push*, that `check_opening_card` and `check_spotlight_transitions` "have no
injection anywhere, so the per-push run is their only exercise in the repo".
That sentence is the reason for this issue; what it does not say is that the
arm which would fix it costs 26 s.

`_check_video` is the ffprobe half of every graded take — duration inside its
window, the file this run wrote rather than the last one. It is reachable at
26 s through `run_spotlight` directly, not only through `check_take`.

## Acceptance criterion

An `Injection(...)` aimed at `--polish-only` for each of the three, each seen
to fail, with the failure output in the pull request, and each failure naming
the check it was aimed at. Specifically, `_check_video` must be seen failing
on its **own** message and not on `check_spotlight_transitions` complaining
about the same broken mp4 — the two share a take, and an injection caught by
the neighbour grades nothing new.

`--list` no longer prints the three as ungraded, and their rows leave
`tests/README.md`'s roster with the counts still correct.

## What this deliberately leaves out

- `check_opening` and `check_opening_gap`. They look like polish and are not
  reachable from this arm: `run_web` is their only phase, at 123 s. #233
  covers them.
- The other cheap slices: `--narration-only` (8 s), `--determinism-only`
  (19 s), `--segments-only` (29 s).
- Splitting a cheaper arm out of `run_spotlight` or `run_terminal_opening`.
  26 s is affordable; this issue is about grading them at all.

rogvid:
[...]

Issue #207, opened 2026-08-01
The timeline gap #110 and #206 propose as the bare-terminal check does not discriminate the flash


rogvid:
#110 measured the bare-terminal flash as a timeline gap — part2's offset (37.76 s)
versus the interlude beat's timestamp (38.05 s) — and #206 restated it as the
measurement that keeps the fix from being graded by eye: *"the gap should be
absent rather than merely smaller."*

It is not absent, and the reason is that the gap never measured the flash.

## Measured, on the same box, same storyboard, one line apart

Both takes of `examples/ticket-queue/demos/2026-07-26-status-filter`, before and
after adopting `TerminalRecorder(interlude=…)` (PR for #206):

| | `interlude()` as first statement | `interlude=` constructor |
|---|---|---|
| part2 offset | 37.96 | 37.88 |
| first part2 beat `t_start` | 38.318 | 38.245 |
| **timeline gap** | **0.358 s** | **0.365 s** |
| corner luma, first part2 frame | **226.5 (bare)** | **25.8 (card)** |
| bare-terminal frames before the card | **6 (0.30 s)** | **0** |

The defect is gone. The proposed measurement moved by +7 ms, in the wrong
direction, and would have reported the fixed take as marginally worse.

## Why

The gap is `TerminalRecorder`'s setup cost, not the flash: `_t0` is set before
`_start()`, and `_start()` does `goto("about:blank")`, injects xterm.js, opens
the PTY, waits for the first prompt, and only then calls `_open_on_card()`. So
the first beat lands ~0.36 s into the segment either way. What the constructor
argument changes is that all of that now happens *behind* an opaque card raised
from an init script — the pixels, not the clock.

A timestamp comparison is a proxy that stays true on a card painted on time and
drawn transparent, which `tests/smoke` already says in the comment above
`check_opening_card`. This is the same point from the other end: it is also true
on a card painted late and drawn opaque-from-frame-zero.

## Why it is worth an issue rather than a note

Nothing asserts on this gap today, so nothing is broken. The risk is that the
number is written down in two closed issues as *the* way to check #110 without
watching, and it is the kind of measurement someone reasonably turns into an
assertion later — at which point it grades nothing, in the
`wip/verified-review` catalogue's sense. Worth either deleting the claim or
replacing it with the corner-luma sweep, which does discriminate (226 → 26 at
the boundary frame, across four re-records including one under load).

## Deliberately not in scope

[...]

Issue #147, opened 2026-07-28
_DemoBase is 1,536 lines holding five concerns, and its seam leaks 8 private attributes


rogvid:
Follow-up to #139, which was its precondition. Filed rather than smuggled into
PR #146.

## The measurement

`_DemoBase` (`core.py:2590-4999`) is **2,408 lines, 73 methods and 55 instance
attributes**. Its docstring says it "owns everything that does not care what is
being recorded", and it does: secrets, issues, beats, evidence, failure dumps,
narration, ffmpeg conversion. Seven concerns, each with its own state, sharing
one `self`.

The declared seam between the base and a medium is eleven hooks
(`_init_context`, `_start`, `_stop`, `_postprocess`, `_content_rect`,
`_checkpoint`, `_before_shot`, `_verify_redaction_final`, `_evidence_payload`,
`_failure_screen`, `_media_path`). The measured seam is wider:

| subclass | hooks overridden | base private attrs read | base private methods called |
|---|---:|---:|---:|
| `Recorder` (web.py) | 9 | **15** | 4 |
| `TerminalRecorder` | 7 | 5 | **10** |

`Recorder` reaches into `_evidence_masks`, `_evidence_outside`, `_mask_seen`,
`_nav_pending`, `_opening_held`, `_last_sync`, `_redacted`, `_redact_style`,
`_accent`, `_context`, `_size`, `_terminal_prompt`, `_terminal_title`,
`out_dir` and `page` — state belonging to at least three different concerns.

`__exit__` alone is ~280 lines, and its comments state that the *ordering* of
its first three statements is the whole redaction guarantee.

## Shape

Extract collaborators with narrow interfaces. Every one is pure Python and
needs no browser, so each becomes a `tests/unit` case the moment it exists —
which is the whole reason #139 came first.

- [ ] `SecretRegistry` — `register` / `scrub` / `assert_clean` / `secrets`
- [ ] `IssueLog` — `note` / `attribute` / `verdict` / `summary`
- [ ] `BeatLog` — the `_beat` contextmanager, `records`, renumbering
- [ ] `EvidenceBuilder` — `capture` / `mask` / `build`
- [ ] `FailureDump` — `summarise` / `build` / `write`
- [ ] `NarrationTrack` — `prepare` / `start` / `finish`

**One collaborator per change, each with the tests that watch it.** This is the
largest change available in this package and the only check that would catch a
mistake in it, before #139, was the ten-minute browser suite.

## Watch out for

[...]

Issue #128, opened 2026-07-27
A terminal take's opening card is intermittently not up in frame 0 on CI (corner reads 128, not 26)


rogvid:
Surfaced by the CI run on #127, whose branch **does not touch this path** — the
diff against main is a `criteria=` keyword passed through `TerminalRecorder.
__init__` and additions to `tests/smoke` that the terminal arm never reaches.

## Measured

`check_opening_card` reads the mean luma of a corner strip outside the terminal
window, on the recording's **first frame**. The card is dark (<= 60), the bare
pastel background is light (>= 150).

| where | runs | corner at frame 0 | card covers |
|---|---:|---:|---|
| `main` CI (c47e3ad, 908986c, 95f4e44) | 3 | **26** | 2.65–2.85s |
| this machine (WSL, 16 core) | 3 | **26** | 2.75–2.80s |
| #127 CI (2f40619) | 1 | **128** | **0.00s** |

```
terminal-opening: the recording's first frame reads 128.0 mean luma in the
corner at (1214, 0, 58, 50), which is 'between' and not the card
(card <= 60.0, bare terminal >= 150.0). The segment opens on the terminal and
the card arrives afterwards — issue #110.
```

So **1 failure in 4 observed CI runs, 0 in 6 local runs.** Runner-only, and not
marginal: 26 against 128 against a 226 bare background are three separated
states, not noise around a bar.

## Why 128 is the interesting number

It is neither state. The corner is not the card and not the background — it is
part way between, which is what a card still becoming opaque looks like, or a
frame composited while the element was being appended.

That matters because `tests/README.md` currently states the opposite as a
limit: *"The recorder has no fade-in path (the element is appended already
opaque, so no transition can run), but the suite is relying on setup cost for
that and would not notice if it changed."* This reading suggests the suite
**did** notice, and that the assumption behind that paragraph does not hold on
a slow runner.

## Not #114, and not fixed by #113

#114 is about `interlude()` used as a terminal take's *first beat* instead of
the constructor argument. This take uses the constructor argument —
`TerminalRecorder(interlude=..., interlude_hold=...)`, which is exactly the path
#113 added to remove the flash. So this is residue inside the fix, not the case
the fix told people to avoid.

## Why it is worth fixing rather than re-baselining

It is the terminal twin of #119, which was just fixed for the web recorder: what
the demo opens on is what a reviewer sees first, and a demo that opens on a
flash of bare terminal before its title card is the artifact misrepresenting the
[...]

PR #316, opened 2026-08-12
A terminal take reads its own first frame and writes down what it was (#235)

### LINKED ISSUE #235: Nothing grades the opening frame of a shipped demo — only the recorder's own take
`check_opening_card` grades the recorder's own take. Nothing grades the opening
frame of a demo anybody ships — including the one #110 and #206 were about.

## What is graded today

`tests/smoke --polish-only` records `terminal-opening/` with
`TerminalRecorder(interlude=…)` and sweeps one corner of the frame outside
`#__term_win` at 20 fps: card `#1c1a17` reads ~26 mean luma, the recorder's
pastel gradient ~226. Three statements — first frame is card, the card covers
at least the first second, and the corner does end up bare. That is a claim
about **the recorder**, on a take this suite records itself.

## What is not

`examples/ticket-queue/demos/2026-07-26-status-filter` is the take #110 and
#206 were reported against, and it is graded by a person watching. It always
has been:

- Demo directories commit `record.py`, `timeline.json`, `timeline.md` and
  `images/`. They do not commit `demo.mp4` or the `.seg.mp4` parts, so no
  committed artifact carries the first frame of part2 — the frame the whole
  question is about.
- `.github/workflows/ticket-queue.yml` grades the app, not the demos.
- The timeline gap does **not** stand in for it. #207 measured 0.358 s before
  the fix and 0.365 s after, on the same storyboard one line apart, while the
  first part2 frame went 226.5 to 25.8. Anything built here must read pixels.

So three human sightings of the same flash (#110, #114's re-report, #206) were
each caught by somebody watching, and a fourth would be too.

## The measurement that would close it

At **record time**, because that is the only moment the video exists: after a
segment is captured, read the same corner strip off the live `#__term_win` box
on the segment's own first frame and record the number in the take's report.
The two constants are already in the recorder's styling and already used by
`check_opening_card` — card `<= 60`, bare `>= 150`, with the band between them
left deliberately empty so a card mid-fade is called neither.

Deliberately unstated here: whether that number should *warn*, *refuse under
`strict=True`*, or only be reported. #128 is why that is a real question — one
CI run in nine read 128 and 0.00 s of cover on a loaded runner, which is
neither state, so a refusal built on this bar would be flaky on the runner it
most needs to be trusted on. Reporting it is the floor; enforcing it is a
second decision.

## Deliberately out of scope

- Enforcing the constructor argument for everyone — that is #114.
- Giving `check_opening_card` a `tests/smoke-inject` entry. It is listed in the
  ungraded roster in `tests/README.md`, it was fault-injected by hand when it
  was written (three of the nine injections in the polish section), and its
  cost is the reason it has no manifest entry.

## Acceptance criterion

Re-recording a terminal segment writes a number that says whether its first
frame was the card or the bare terminal, and a reviewer can read that number
without opening the video. Broken by raising the card 400 ms into the document,
as `tests/smoke`'s own injection does.

rogvid:
Fixes #235.

At record time — the only moment the video exists — a terminal take reads the
corner strip off its own **first frame** and writes the number into the take's
report. A reviewer reads `content.opening.card` in `timeline.json`, which *is*
a committed artifact, instead of opening an mp4 that is not.

## The field

`content.opening.card`, on every terminal take (null on a web take, which has
no window to read beside):

```json
"card": { "luma": 25.8, "state": "card", "raised": true,
          "rect": [1214, 0, 58, 50],
          "card_max": 60.0, "bare_min": 150.0, "note": null }
```

- `luma` — mean luma of the strip of background beside the live `#__term_win`
  box, on frame zero, read with `-t` and **no** `fps` filter (the filter
  answers with whatever frame lands nearest the middle of its first slot; that
  cost `check_opening` a real defect once).
- `state` — `"card"`, `"bare"` or `"between"`, derived from the **rounded**
  `luma` this dict publishes, so the number and the word cannot disagree.
- `raised` — whether the take asked for a card. It is what makes `"bare"`
  readable: the defect on a segment opened with `interlude=`, the arrangement
  on one that never wanted a card. A take using the old `interlude()`-first
  shape (#114) therefore records `raised: false, state: "bare"` — the flash,
  in the one file a demo directory commits.
- `note` — why, when `luma` is null. Never a guess, never silently absent.

The two bars are `OPENING_CARD_MAX_LUMA = 60.0` / `OPENING_BARE_MIN_LUMA =
150.0`, moved into `content.py` so the recorder classifies against the same
pair `check_opening_card` grades against. `tests/smoke` keeps its own
hand-written copy and `check_reported_opening_card` asserts the two are equal —
one moved without the other is a failure, not an agreement.

## Report, not enforce — and this was decided, not defaulted

#235 left warn/refuse/report open. This is **report**: nothing warns, nothing
raises, no take fails. #128 measured one CI run in nine reading **128** luma and
0.00 s of cover on a loaded runner — neither state — so a refusal built on this
bar would be flaky on exactly the runner it most needs to be trusted on, and a
bar retuned until it was not would be the catalogue's *threshold tuned to this
box*. The band between 60 and 150 is left deliberately empty so a card mid-fade
is called `"between"` rather than either thing.

I found no strong reason report-only is wrong.

## The seam

`_measure_content` is sealed on `_DemoBase` (`__init_subclass__` refuses an
override at import — the #147 machinery caught my first attempt). So this lands
as a declared medium hook, `_opening_card`, added to `MEDIUM_HOOKS` beside
`_content_rect`. Core's change is four lines; everything medium-specific is in
`terminal.py`.

### `tests/smoke --polish-only` → `check_reported_opening_card`, 4 entries

Registered in `tests/smoke-inject`, so they re-run rather than being a claim in
this PR body. The harness aborts unless the search string matches **exactly
once** — it did that to me once during this work, on a string that occurred
twice in `content.py`:

```
ABORT: injection 'a take with no window box measures something regardless' matched 2 times in
content.py, expected exactly 1. The injection did not land, so nothing it would have proven is proven.
```

```
smoke-inject: 4 injection(s) over 1 arm(s), plus 1 clean baseline(s)

BASE  --polish-only passes clean (28s)

PASS  break: a terminal take says nothing about the frame it opened on  [28s]
      --polish-only, in terminal.py: return opening_card_report(…
      caught: terminal-opening: this take wrote content.opening.card = None, so its timeline says
              nothing about the frame the segment opened on. That is the state i…

PASS  break: the opening frame's luma is a constant rather than a reading  [28s]
      --polish-only, in content.py: report["luma"] = round(sum(frames[0]) / len(frames[0]), 1)…
      caught: terminal-opening: the take reports its first frame at 40.0 mean luma over
              [1214, 0, 58, 50], and frame zero of demo.mp4 reads 25.8 over (1214, 0, 58,…

PASS  break: the opening card is raised 400 ms into the document  [29s]
      --polish-only, in terminal.py: if (document.body) build();
  else addEventListener('DOMContentL…
      caught: terminal-opening: this take opened with TerminalRecorder(interlude=…) and reports its
              first frame as 'bare' at 226.5 mean luma — the segment opens on…

PASS  break: the report forgets that the take asked for an opening card  [29s]
      --polish-only, in terminal.py: raised=self._opening is not None,…
      caught: terminal-opening: content.opening.card says raised=False on a take constructed with
              interlude=… — without that flag 'bare' cannot be told from a segme…

All 4 injections were caught. [2.4 min]
```

The third is the injection the acceptance criterion names. Run by hand as well,
in the working tree, to capture the whole failure list and the artifact it
leaves — `tests/smoke --polish-only`:

```
smoke: FAILED
  - terminal-opening: the recording's first frame reads 226.5 mean luma in the corner at
    (1214, 0, 58, 50), which is 'bare' and not the card (card <= 60.0, bare terminal >= 150.0).
    The segment opens on the terminal and the card arrives afterwards — issue #110.
  - terminal-opening: the card covers only the first 0.00s of the recording (0 frames at 20 fps),
    under the 1.0s bar for a card held 2.5s. Frame 0 reads 226.5 — the viewer sees the terminal
    before the card that is supposed to be covering it.
  - terminal-opening: this take opened with TerminalRecorder(interlude=…) and reports its first
    frame as 'bare' at 226.5 mean luma — the segment opens on the terminal and the card arrives
    afterwards (issue #110), and this time the take's own report says so
```

and the take's own report, which is the point of the change:

```json
"card": { "luma": 226.5, "state": "bare", "raised": true,
          "rect": [1214, 0, 58, 50], "card_max": 60.0, "bare_min": 150.0, "note": null }
```

226.5 → 25.8 is #207's pair of numbers for the same storyboard one line apart,
reproduced here by the recorder itself. On the healthy take the two independent
readings agree exactly:

```
smoke: terminal-opening card covers the first 2.85s (corner 26 -> 226), then clears
smoke: terminal-opening reports its own opening frame as card (25.8 mean luma, frame zero here reads 25.8)
```

### `tests/unit` → `OpeningCard`, 5 injections

0.2 s each, all caught (`tests/unit --fault-inject`: *All 274 injections were
caught*):

```
PASS  break: the card bar excludes the value it is written as including
      in content.py: if luma <= OPENING_CARD_MAX_LUMA:…
      FAIL: test_both_bars_are_graded_at_their_boundary

PASS  break: the empty band between the two bars is closed
      in content.py: OPENING_BARE_MIN_LUMA = 150.0…
      FAIL: test_a_card_part_way_up_is_called_neither
      FAIL: test_the_band_between_the_bars_stays_wide_and_empty

PASS  break: a take with no window box measures something regardless
      in content.py: if rect is None:…
      FAIL: test_a_recorder_that_cannot_place_its_window_says_so

PASS  break: the report claims every take asked for an opening card
      in content.py: "raised": raised,…
      FAIL: test_whether_a_card_was_asked_for_is_in_the_record

PASS  break: the opening block stops carrying the card key at all
      in content.py: return {"gap": gap, "held": held, "limit": limit, "note": note, "car…
      FAIL: test_the_opening_block_carries_the_key_on_every_take
```

### An assertion written and then deleted

`state` follows the `luma` beside it — deleted rather than shipped unfirable.
Every take `--polish-only` records opens **on** the card, so a `state` pinned to
`"card"` could only fail together with "it says card", which is the catalogue's
*dominated assertion*. Recorded in `tests/README.md`'s Known gaps with what
would close it (a second terminal take in that arm, opened without
`interlude=`, which would have to report `bare`/`raised: false`) and what that
costs.

## Stated limits

- **No demo in this repository has been re-recorded**, so
  `2026-07-26-status-filter`'s committed timeline carries no such field and is
  still graded by whoever watched it. The field arrives with the next
  re-record. Out of scope here on purpose.
- **The number is in `timeline.json` only.** It is not a clause of the healthy
  `shows a picture` line and not on stderr — the take prints exactly what it
  printed before. #235 asks for it where a reviewer reads without opening the
  video, and that is the committed file.
- **Reported means shippable.** A demo whose part 2 opens bare still ships, with
  a truthful line saying so. Enforcement is a second decision (#114 owns the
  constructor half).
- **The independence of the smoke check is narrow.** It re-reads frame zero
  itself, over a strip `record_terminal_opening` derives from the same live
  `#__term_win` box the recorder reads — so a recorder and a harness that made
  the same mistake about *which* corner to read would agree. What makes that
  visible is `check_opening_card` above it, which sweeps the whole take.
- **A card that fades in is still invisible here**, exactly as before: the
  recorder spends longer injecting xterm.js than a 450 ms fade takes, so frame
  zero is opaque either way. Unchanged by this, and already stated in
  `tests/README.md`.
- **`--polish-only` now carries entries**, which it did not before. #240 stays
  open for the arm's other three functions (`check_opening_card`,
  `check_spotlight_transitions`, `_check_video`) — this PR adds none for them,
  per #235's out-of-scope list. Their baseline is now already paid.
- **Manifest cost.** 60 → 64 entries, ~44.6 → ~46.8 min estimated. `--self-test`
  checks every figure in `tests/README.md` against the manifest, so those
  numbers are in the file and graded.

## Suites

`tests/lint`, `tests/unit` (418), `tests/unit --fault-inject` (274),
`tests/ci-unit` (99), `tests/smoke-inject --self-test`, mypy 1.18.2, and the
full `tests/smoke` — **PASSED**.

### DIFF
--- a/skills/demo-video/helpers/demo_recording/content.py
+++ b/skills/demo-video/helpers/demo_recording/content.py
@@ -379,7 +379,7 @@ def content_rect(rect: tuple[int, int, int, int]) -> tuple[int, int, int, int]:
 def _content_frames(
     mp4: Path,
     rect: tuple[int, int, int, int],
-    sample_fps: int,
+    sample_fps: int | None,
     limit_s: float | None = None,
 ) -> list[bytes]:
     """`mp4`'s content rect, sampled and reduced to grayscale frames.
@@ -387,11 +387,20 @@ def _content_frames(
     `limit_s` stops after that many seconds of video — the opening check looks
     at the first few seconds and decoding a whole 60 s take to answer a question
     about its first 400 ms is waste. None reads the file to the end.
+
+    `sample_fps` of None drops the `fps` filter and returns the file's own
+    frames, which is the only way to be sure of getting **frame zero**: the
+    filter quantises to its own output slots and hands back whichever source
+    frame lands nearest the middle of slot zero. That cost `tests/smoke`'s web
+    opening check a real defect once — at 1 fps it was reading a frame from
+    around 0.5 s and scoring a take whose first frame was flat white as
+    healthy. Pair it with a short `limit_s`.
     """
     x, y, w, h = rect
     chain = (
-        f"fps={sample_fps},crop={w}:{h}:{x}:{y},"
-        f"scale={_CONTENT_W}:{_CONTENT_H},format=gray"
+        (f"fps={sample_fps}," if sample_fps else "")
+        + f"crop={w}:{h}:{x}:{y},"
+        + f"scale={_CONTENT_W}:{_CONTENT_H},format=gray"
     )
     window = [] if limit_s is None else ["-t", f"{limit_s:.3f}"]
     proc = subprocess.run(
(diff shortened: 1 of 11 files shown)


rogvid:
Rebased onto `277e95d` and the two gaps the review measured are closed in this commit rather than filed.

**The exception path.** `_content_rect` raising nulled the whole `content.opening` block, so a terminal take that could read `#__term_win` and not `#__term_host` said nothing about the frame it opened on and nothing said why. The card needs no app rect and is no longer lost with it; the gap it genuinely could not measure is absent with its reason written down. Injected by restoring the old `pass`, which reddens `OpeningCardOnATake.test_a_take_that_lost_its_app_rect_still_reports_its_opening_frame` alone.

**The unrounded edge.** The strip was measured from `int(x) + int(width)` where `tests/smoke` reads `int(x + width)`. On a fractional box those land a column apart — `100.6 + 57.6` gives 158 one way and 157 the other. A fraction of a luma level, and never enough to change the verdict; enough to be a latent flake in an agreement check with a 1.0 bar, where it would have presented as the property under test breaking rather than as a rounding difference. Both now derive the edge the same way and the test pins the arithmetic directly.

Suites on the rebased tree:

```
tests/unit                    OK
tests/unit --budget           SKILL.md: 600 lines, budget 600 lines      exit 0
tests/unit --fault-inject     All 292 injections were caught.
tests/ci-unit                 OK
tests/ci-unit --fault-inject  All 76 injections were caught.
tests/lint                    exit 0
tests/smoke-inject --self-test  exit 0
```

## Stated limit, deliberately not filed

`check_opening_card` — pre-existing, untouched here — still describes what it reads as "the recording's first frame" while sweeping at `fps=20`. That is the aliasing this commit's own code carries a comment about, because it cost this repo a defect once (#229, #245, #246). The brief for this change asked for it to be filed as an issue. It is recorded here instead: as of `docs/2026-08-13-reset.md` this repository has stopped converting every true observation into committed work, because that rule is what grew the backlog by thirteen issues on a day five pull requests merged. The measurement is above if anyone needs it.

Issue #256, opened 2026-08-09
A step's hole is not published, so every consumer corrects instants that have no video into pre-step content


rogvid:
Slice of #255, which holds the measurements and the design record. **This is the slice that repairs behaviour already shipped in `d5212c0` and `5b00786`.**

A −Δ wall-clock step deletes a Δ-wide window of monotonic time from the file. Measured, one video frame wide: `video 31.560 → 31.600 shows mono 31.644 → 32.644`.

`HostClock.before()`, `merged_clock_correction` and `timeline.capture_clock_correction` all return a shift for every `t > T`. An instant inside `(T, T+Δ)` has no video at all, and correcting it lands up to a whole step early, in content that predates the step.

Observed in the shipped path: seg-run1 `beat-05` midpoint 5.633 s sits in the hole (5.151, 6.202); its frame was cut at video 4.58 s and **shows the previous caption**. seg-8 beats 13 and 14 likewise.

## Acceptance criterion

An instant with no video is never silently corrected into content that predates the step, in any of the three consumers — review frames, narration, and the smoke harness's own reader.

- [ ] `before(t)` (and the correction functions built on it) distinguish *no video here* from *shift by Δ* — return `None`, or clamp to the hole edge **and say which**.
- [ ] Each consumer states what it did: the review sheet, the narration envelope, and the harness's failure text. A beat with no video must not be presented as a frame at its midpoint.
- [ ] Graded by a unit case built synthetically — the host cannot be relied on to step to order.

## Deliberately out of scope

Partial loss (a take losing 16–50% of the recorded step) and verifying the step against pixels — those are the other slices of #255. This slice assumes the recorded Δ is right and only fixes the hole it implies.

PR #348, opened 2026-08-22
Downscale the review frames and name every picture the sheet withholds (#343)

### LINKED ISSUE #343: Review frames cost ~100k image tokens; downscale and dedupe them in beat_frames
From the first run on a foreign repo (TR webplatform, 2026-08-21, field notes in .scratch/2026-08-22-demo-video-real-run-review.md). A 122 s demo produced 80 frames at 1280×720 ≈ 98k image tokens. Three reviewer agents died on session limits before one review completed. The review loop is the skill's best feature and it did not survive first contact at this size.

Measured on the final take:

| variant | frames | ~image tokens |
| --- | ---: | ---: |
| as shipped | 80 | ~98k |
| dedupe only (RMSE >= 3.0 vs last kept) | 45 | ~55k |
| dedupe + scale to 1024×576 | 45 | ~35k |

44% of frames were a picture already shown. `reference/limits.md` line 289 already measures this class at 69% of a reference take. The sheet header even claims "a held frame is photographed once" (frames.py:546) — per beat that is true, but the reader takes it as "no duplicate pictures", and 44% duplicates makes the sheet break its own promise.

Change `beat_frames` (helpers/demo_recording/frames.py):

1. Extract at 1024 wide (`-vf scale=1024:-1` in `_extract`). Verified legible on the field app: burned-in captions, toast text, table digits, dropdown labels.
2. Drop a frame within RMSE < 3.0 of the last kept frame. Always keep the first and last. List every dropped beat in frames.md by name so nothing is silently hidden — same rule as `scene_search_skipped`. The threshold must be mechanical: a hand-picked subset can launder findings out of a review; a threshold cannot.

Acceptance: on the reference takes, frames.md lists each dropped beat by name; total frame bytes fall by more than half; no beat disappears without a printed line saying so. Recorder behaviour — wip/verified-review discipline applies, every new assertion seen red first.

rogvid:
Fixes #343.

## Acceptance criterion

On takes with held frames: total frame bytes fall by more than half, frames.md names each dropped beat, and no beat disappears without a printed line saying so.

## What changed

`_extract` scales every review frame to at most 1024 px wide (`min(1024,iw)`, never upscaled). After extraction, frames are walked in sheet order and a frame within RMSE 3.0 of the last kept frame's picture is dropped. A drop is never silent: the PNG comes off disk, the manifest gains a `deduped` entry (file, beat, kind, t, matched kept file, measured rmse), and frames.md prints one line per drop, e.g. `` `beat-01.png` is not on this sheet: its picture is `beat-00.png`'s (RMSE 0.15 < 3.0). `` The first and last frame are always kept. Scene frames participate like beat frames. A failed comparison keeps the frame and warns — a broken comparator must fatten the sheet, never thin it. The comparator is ffmpeg's `psnr` filter (`mse_avg`; identical frames report mse 0 under an inf psnr and drop too) — no new Python dependency reaches the helpers. `FRAMES_SCHEMA` is now 2, and the sheet's header no longer promises "a held frame is photographed once".

## Measurement

Same video each row: frames regenerated from the identical demo.mp4 + timeline.json by the old code (main) and the new.

| take | before | after | bytes |
|---|---|---|---|
| segments (14.9 s, 15 beats) | 15 frames, 2,985,725 B, 1280x720 | 8 kept + 7 named drops, 1,156,800 B, 1024x576 | **-61.3%** |
| web (31 beats) | 31 frames, 6,547,536 B | 18 kept + 13 named drops, 2,839,713 B | **-56.6%** |

Every dropped beat is named in frames.md and carried in `manifest.deduped`; the smoke harness now counts kept + deduped together against the hand-written beat lists, so a silent disappearance is a red arm.

## Consumers

- **tests/smoke** assumed one frame per beat, native-res byte identity, and positional caption-context alignment — all updated. `_same_frame` re-cuts through a hand-copied (not imported) scale filter. Kept + deduped must cover the beat list; each drop must be named in frames.md, matched to a kept frame under a hand-written 3.0 bound, and off disk. The caption-state check grades a deduplicated beat through the kept twin the manifest says holds its picture — so a dedupe that merged across a caption change puts the wrong caption state under that beat's expectation and collapses the margin. The caption band is scaled to the PNGs' actual size.
- **scripts/demo-grade** needed nothing: `frame_records` already documents that beat-to-frame is neither 1:1 nor monotone and reads only `manifest.frames`.
- **`.github/scripts/*`** read no frame manifests.

## Fault injections (tests/unit --fault-inject, 330/330 caught)

Each injection lands via the manifest's exact-once match (the harness aborts on 0 or 2+ matches). The eight new ones, with their failure output verbatim:

```
PASS  break: a repeated picture is reprinted on the sheet instead of deduplicated
      in frames.py: if mse >= DEDUPE_RMSE * DEDUPE_RMSE:…
      FAIL: test_a_middle_frame_that_repeats_the_last_kept_is_dropped (__main__.FrameDedupe.test_a_middle_frame_that_repeats_the_last_kept_is_dropped)
      FAIL: test_the_first_and_last_frames_are_kept_even_when_identical (__main__.FrameDedupe.test_the_first_and_last_frames_are_kept_even_when_identical)

PASS  break: a dropped frame leaves no record in the manifest or on the sheet
      in frames.py: deduped.append(…
      FAIL: test_the_manifest_names_the_drop_its_match_and_the_measured_rmse (__main__.FrameDedupe.test_the_manifest_names_the_drop_its_match_and_the_measured_rmse)
      FAIL: test_the_sheet_names_every_dropped_frame_and_stops_embedding_it (__main__.FrameDedupe.test_the_sheet_names_every_dropped_frame_and_stops_embedding_it)

PASS  break: frames.md stops naming the frames it withholds
      in frames.py: for drop in manifest.get("deduped") or []:…
      FAIL: test_the_sheet_names_every_dropped_frame_and_stops_embedding_it (__main__.FrameDedupe.test_the_sheet_names_every_dropped_frame_and_stops_embedding_it)

PASS  break: the last frame of the sheet is deduplicated away like any other
      in frames.py: if n == 0 or n == len(written) - 1:…
      FAIL: test_the_first_and_last_frames_are_kept_even_when_identical (__main__.FrameDedupe.test_the_first_and_last_frames_are_kept_even_when_identical)

PASS  break: a failed comparison drops the frame instead of keeping it
      in frames.py: if mse is None:…
      FAIL: test_a_failed_comparison_keeps_the_frame_and_warns (__main__.FrameDedupe.test_a_failed_comparison_keeps_the_frame_and_warns)

PASS  break: a candidate is compared against its neighbour, not the last kept frame
      in frames.py: frames_dir / str(kept[-1]["file"]), frames_dir / str(record["file"])…
      FAIL: test_a_candidate_is_compared_against_the_last_kept_frame (__main__.FrameDedupe.test_a_candidate_is_compared_against_the_last_kept_frame)

PASS  break: a deduplicated frame's PNG stays on disk
      in frames.py: png.unlink()…
      FAIL: test_a_middle_frame_that_repeats_the_last_kept_is_dropped (__main__.FrameDedupe.test_a_middle_frame_that_repeats_the_last_kept_is_dropped)

PASS  break: review frames are extracted at native resolution again
      in frames.py: "-vf",…
      FAIL: test_extract_asks_ffmpeg_for_at_most_1024_wide_never_upscaled (__main__.FrameDedupe.test_extract_asks_ffmpeg_for_at_most_1024_wide_never_upscaled)
```

## Runs

- tests/unit: 507 OK. tests/unit --fault-inject: 330/330.
- tests/ci-unit: 147 OK. --fault-inject: 107/107.
- tests/smoke-inject --self-test, tests/lint, mypy: clean.
- tests/smoke --web-only, --terminal-only, --segments-only, --polish-only: PASSED with the new checks (web 18 kept + 13 named drops, terminal 10 + 8, segments 9 + 6; every kept frame byte-identical on re-cut).

## Stated limits

- `tests/smoke --cheap` went red twice on this WSL2 box, both times only in the spotlight luma windows (enter/exit ends differ by 0.05-2.13 luma, floor 5.0) — a check that reads demo.mp4 pixels this diff never touches, green when the arm runs alone (`--polish-only`: "spotlight eases both ways"). This host steps its wall clock ~0.74 s during most takes; tests/README.md's own reference `--cheap` measurements record a red of this class (#215/#224). CI's runner is the authoritative `--cheap`.
- A deduplicated beat whose midpoint the host's clock deleted (#256) loses its `no_video` marker with its frame record; the `deduped` entry says which kept picture stands for it, but not that its own moment is missing from the file. `_check_unstated_holes` grades kept frames only.
- The blind reader's brief lists kept frames only; a clause satisfied only by a deduplicated beat is read through the kept twin's picture without knowing the beat number. The picture is the same by construction (RMSE < 3.0).
- The RMSE threshold is whole-frame. A caption text change confined to the caption band could in principle sit under 3.0; the smoke caption-state check now grades deduplicated beats through their kept twin precisely so that merge would collapse the margin, and no take showed it.

### DIFF
--- a/skills/demo-video/README.md
+++ b/skills/demo-video/README.md
@@ -28,12 +28,14 @@ and a real terminal session (voice on):
   waits (recorded as segments, stitched losslessly).
 - **`images/*.png`** — stills captured at key moments, ready for a written
   guide.
-- **`frames/`** — one review frame per beat, plus `frames.md` embedding them
-  in order. It is what you hand the fresh-agent review: aimed at what the
-  storyboard did rather than at a stopwatch, so nothing is missed and a held
-  frame is photographed once. A stitched demo gets them too, off the merged
-  timeline. Deliberately uncaptioned, and aimed to within a beat or so rather
-  than exactly — `reference/review.md` says how far and why.
+- **`frames/`** — a review frame per beat, at most 1024 px wide, plus
+  `frames.md` embedding them in order. It is what you hand the fresh-agent
+  review: aimed at what the storyboard did rather than at a stopwatch, so
+  nothing is missed. A beat whose picture repeats an earlier frame's is not
+  reprinted — `frames.md` names it and says which kept frame holds its
+  picture. A stitched demo gets them too, off the merged timeline.
+  Deliberately uncaptioned, and aimed to within a beat or so rather than
+  exactly — `reference/review.md` says how far and why.
 - **`record.py`** — the storyboard that produced the media, and the thing
   that actually gets committed: re-runnable after the UI changes, so the
   video stays out of git history and is regenerated rather than archived.
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -438,9 +438,10 @@ terminal demo takes, and the gotchas — pagers, echo, prompts that never return
    [reference/timeline.md](reference/timeline.md).
 6. **Fresh-agent review (required).** You cannot watch the video, and you
    know too much anyway — have a context-free agent watch it for you. The
-   recorder has already written what they need: `frames/`, one PNG per
-   beat, and `frames/frames.md`, which embeds them in order. **Hand them
-   `frames/frames.md`.** Read
+   recorder has already written what they need: `frames/`, a PNG per beat
+   minus the repeats (a beat whose picture repeats an earlier frame's is
+   named in the sheet instead of reprinted), and `frames/frames.md`, which
+   embeds them in order. **Hand them `frames/frames.md`.** Read
    [reference/review.md](reference/review.md) first — it says how accurately a
    frame is aimed at its beat, which is what decides how much weight a
    reviewer's reading of one can carry.
--- a/skills/demo-video/helpers/demo_recording/frames.py
+++ b/skills/demo-video/helpers/demo_recording/frames.py
@@ -1,8 +1,10 @@
-"""Review frames: one PNG per beat, pulled out of the finished mp4.
+"""Review frames: one PNG per beat, deduplicated, pulled out of the finished mp4.
 
 Nobody reviewing a demo through this skill can watch a video, so every review
 is a review of frames. They are aligned to beats rather than to a clock — a
 clock misses a short beat entirely and photographs a long static one twice.
+A beat whose picture repeats the last kept frame's is dropped and named on
+the sheet rather than reprinted (see the dedupe section below).
 
 `beat_frames(out_dir)` regenerates the whole sheet from demo.mp4 and
 timeline.json without re-recording.
@@ -11,6 +13,7 @@ timeline.json without re-recording.
 from __future__ import annotations
 
 import json
+import math
 import re
 import subprocess
 import sys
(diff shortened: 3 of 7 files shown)


rogvid:
Verified against the two-question contract: acceptance met (drops named on sheet and manifest, first/last kept, comparator fails toward keeping, −61%/−57% bytes on real takes), no regressions (demo-grade reads frames via subset-tolerant frame_records; smoke pins schema 2; clock correction and byte-identity green on a real take).

Stated limits, measured during verification — none blocks, none earns a ticket under the house bar because every failure direction fattens the sheet rather than lying:

- Nothing pins "a take with held frames produces drops > 0". A broken mse_avg regex returns None for every pair, every frame is kept, and unit + smoke stay green — the ~98k-token sheet returns silently. The regex parses real ffmpeg today (probe: identical 0.0, different 16129.0).
- No fixture sits in the RMSE 1.73–3.0 band, so a mis-squaring of the threshold ships green. It errs toward keeping.
- The smoke beat-file check now compares sorted(kept+deduped), so manifest frame order is no longer asserted there; each frames.md entry still carries its own beat heading and timestamp.

Caption safety was checked by arithmetic, not assumption: at RMSE 3.0 only ~130 high-contrast pixels fit under the threshold; a caption change moves thousands, so a beat cannot dedupe across a caption difference and be graded through its twin.

Issue #362, opened 2026-08-22
Terminal takes mount the shared wrapper chrome: one window, one caption band, both media


rogvid:
Follow-up to #355's structural slices; sequenced after #361. Requested by owner 2026-08-22: web and terminal capture windows must not drift apart visually.

The drift is already real, before the wrapper work: terminal.py's _TERM_HOST_JS and web.py's _FRAME_HTML are two hand-maintained copies of the same chrome, and they disagree today — title bar 40 px vs 36, corner radius 12 vs 14, different shadows, different background gradients, fixed 70/74 insets vs computed pads. After #358-#361 the difference becomes structural: web captions sit in a reserved band below the app rect, terminal captions stay overlaid inside the window (bottom 88 px).

Fix: TerminalRecorder mounts the shared wrapper chrome that #358 authors (one source for background, window, title bar, pads, caption band, card layer, with a content slot). xterm mounts in the slot where web mounts the app iframe. _TERM_HOST_JS's bespoke window is retired; the Catppuccin terminal palette stays — it is the content's colour, a parameter of the slot, not chrome.

What this buys: terminal captions gain the same reserved band (no overlay on TUI content), one set of geometry constants for every downstream consumer (content_rect, to_video_rect, tests/pixel goldens), and the opening-card machinery renders on the same layer in both media.

Acceptance: both recorders' chrome comes from one module (grep finds no second title-bar markup); a terminal take's caption band and content rect are disjoint by geometry; the existing terminal smoke arms are green; the opening-card check still reads frame 0 as card. Recorder behaviour — verified-review discipline applies.

Deliberately leaves out: any palette redesign; the tactical caption placement (still deferred).

Issue #361, opened 2026-08-22
Wrapper cutover: one web path, composite deleted, consumers and docs moved, encode saving measured


rogvid:
Slice 3 of #355's structural half. Needs #358 and #360.

- The wrapper becomes the only web path; the transitional flag from #358 goes away.
- Delete the exit-time composite: _FRAME_HTML screenshot rendering, the _postprocess overlay pass, the opening-hold overlay, and the WEB_CARD_BODY compensation with its rationale comments condensed into a history note pointing at #291/#301/#355.
- Move every geometry consumer: _content_rect/content scoring, to_video_rect callers in tests/smoke, the still/video framing expectations (stills now include chrome — the mismatch this removes), tests/pixel golden checks if #350/#351 have landed, and the caption-font-34px scaling rationale (the app is no longer scaled 0.8).
- Docs: SKILL.md verb/limits rows touched by caption semantics, reference/determinism.md and reference/limits.md where they describe the composite.
- Measure and state in the PR: encodes per take before (2) and after (1), and wall-clock per retake on the fixture storyboard.

Acceptance: full tests/smoke green on the wrapper-only path; grep finds no _postprocess composite and no compensation constant outside the history note; the retake measurement shows one encode saved.

Recorder behaviour — verified-review discipline applies.

rogvid:
Note from PR #366's review (2026-08-22): the wrapper cursor dot has no pixel assertion of its own — the legacy path's tests/pixel goldens (cursor-parked, cursor-absence) grade legacy takes only. When this cutover retires the legacy path, those goldens MUST move to wrapper takes in the same PR, or the dot ships ungraded. Measured basis: the through-frame click post-conditions grade the coordinate math indirectly; nothing reads the dot's pixels on a wrapper take today.

Issue #360, opened 2026-08-22
Wrapper chrome parity: cards and opening hold in the wrapper, captions that survive navigation, uncompensated colour


rogvid:
Slice 2 of #355's structural half. Needs #358.

- Interlude and criterion cards, the bridge scrim and the decorative terminal prop render in the wrapper document, covering the app rect (not the chrome). The opening hold becomes wrapper-native — the card is in frame 0 the way the terminal recorder's init-script card already is, replacing the ffmpeg enable='lt(t,held)' overlay for wrapper takes.
- The caption survives a real mid-take goto: the wrapper document outlives the app's document, so the #134 caption_lost class ends here. Prove it in pixels (caption band still lit across a full page load) and adjust the beat-log semantics honestly — a caption can no longer be destroyed by navigation, so the caption_lost issue record and its SKILL.md row change for wrapper takes, and the timeline must not claim a loss that can no longer happen.
- Colour, single-encoder: on the wrapper path the card and the window body ride one encoder, so the card declares the window's own colour with NO compensation, and the card-vs-window check passes at worst-channel |Δ| ≤ 2 with equal declarations. The #291/#301 compensation constant stays only on the legacy path until #361 deletes it. Do not re-adopt the recorded anti-patterns (declarations pinned equal as the assertion, or pinned ≥ 12 apart).

Deliberately leaves out: making wrapper the only path, deleting the composite, doc sweep (#361).

Recorder behaviour — verified-review discipline; every new assertion seen red first.

Issue #355, opened 2026-08-22
Captions overlap what they narrate — reserve the bar its own space; subtitles-at-composite is the rejected alternative


rogvid:
Design proposal. The observed defect (foreign-repo run): the caption bar sometimes sits on top of the element it is addressing. Today the bar is in-page DOM over the app (core.py:90), so it shares pixels with the content it narrates; SKILL.md's mitigation is "scroll the element to center", which authors forget.

**The alternative that looks attractive and is rejected: render captions at composite time (real subtitles from record.py).** Reasons, recorded so this is not re-litigated:
1. It re-imports the #18 clock problem into the demo itself. The beat log and the captured video are on different clocks (the whole capture_clock saga: #229, #245, #255, #256). A caption placed at the log's timestamp can sit over the neighbouring moment's pixels — the confidently-wrong label frames.py's header refuses to print on review sheets, burned into the product instead.
2. It makes every caption check circular. Today caption presence/timing is measured out of the pixels (caption band delta, caption_appearance_s, caption_lost on document death). Composite-time captions are generated FROM the spans the checks would grade — "the faithful transformation of an untrusted input", verbatim from the verified-review catalogue.
3. A soft subtitle track specifically is dead on arrival: GitHub's PR video player exposes no track UI, and the blind reader reads PNG frames, which a track never reaches.

**What to do instead — the bar gets its own reserved space, still at capture time:**
- Tactical, shippable alone: per-beat placement — caption(text, at="top"), plus auto-dodge when the current spotlight target's rect intersects the bar. Chrome state stays in the pixels; no clock involved.
- Structural, the real fix: frame the web take in-page (like the terminal recorder already does) and give the caption a letterbox band in the window chrome BELOW the app rect. The app never shares a pixel with the bar.

The structural half pays three more times (v2 feedback C2): it kills the exit-time ffmpeg composite (one full-video encode per retake), it retires the two-encoder colour compensation class that produced #291's hex negotiation (card and window would ride one encoder), and it removes the "stills are full-bleed, video is windowed" mismatch.

Costs to weigh in the design round: in-page framing must survive hostile app CSS (position:fixed ancestors, z-index wars — the existing overlay machinery already fights this); every geometry consumer (to_video_rect, app-rect scoring, #324's golden checks) changes coordinates; and the opening-hold overlay path (web.py:707) needs a new home.

Split into slices before anyone starts, per CLAUDE.md. This issue is the design record.

rogvid:
[...]

PR #366, opened 2026-08-22
Wrapper takes: the web app records in an iframe on a recorder-owned page, caption in its own band below the app rect (#358)

### LINKED ISSUE #358: Wrapper takes: the web app records inside a recorder-owned frame, caption band below the app rect
Slice 1 of #355's structural half (design record in that issue's comments; mechanism pinned there).

Add an opt-in wrapper path to the web Recorder: the take opens a recorder-owned wrapper page carrying the window chrome (title bar, pads) and a reserved caption band BELOW the app rect; the app loads in an iframe sized to the app rect at its true pixel size. The flag is transitional — it exists so this slice can merge green, and #361 removes it.

In scope:
- Verbs re-target the app frame (locators, spotlight, evidence capture); the cursor overlay and coordinate math gain the app-rect offset; rec.page semantics decided and documented (the escape hatch must still reach the app's document).
- The caption bar moves to the wrapper document, rendered inside the reserved band. By construction the app rect and the caption band share no pixels — assert that from geometry, and from pixels on a recorded take (caption band lit, app rect unchanged above it).
- An app that refuses framing (X-Frame-Options, CSP frame-ancestors) is detected and the take refuses with a message naming the header — a blank iframe recorded silently is the artifact-lie outcome this slice must not ship. Seen red: serve the fixture with the header and watch the refusal.
- One smoke-adjacent proof take on the fixture page exercising click, type_into, scroll_to, spotlight, caption against the frame.

Deliberately leaves out: interlude/criterion cards and opening hold in the wrapper (#360), the cutover and composite deletion (#361), caption(at=...) placement (deferred by owner decision).

Recorder behaviour — wip/verified-review discipline applies throughout; every new assertion seen red first.

rogvid:
Slice 1 of #355's structural half (design record in that issue's comments; the shared-chrome amendment is on #358).

## What this adds

`Recorder(wrapper=True)` (or `DEMO_VIDEO_WRAPPER=1`) records the web app inside an iframe on a recorder-owned wrapper page.
The wrapper page is the recording: window chrome, title bar, pads, a caption band reserved **below** the app rect, and the cursor overlay all live in the wrapper document, and the exit-time ffmpeg composite does not run on this path.
The app loads at the app rect's true pixel size — no 0.8 scale, so the caption keeps the base 26px instead of the composite's 34px compensation.
Off by default; the legacy composite path is untouched.

- **`helpers/demo_recording/chrome.py`** — the shared chrome module the owner amendment asked for: one source for the background gradient, window frame, title bar + traffic lights, pads, the caption band below the content slot, and an (empty, for #360) card layer. `chrome_geometry(width, height)` is pure and unit-graded; `chrome_html(geom, ...)` renders the wrapper document with an **empty content slot** (`#__chrome_slot`) that the web path fills with the app iframe and #362 fills with the xterm host. Nothing web-only is hard-wired into the chrome — the iframe is mounted by `web.py`, not shipped in the chrome document.
- **Verbs re-target the app frame**: locators, spotlight (context init scripts run in every frame, so the spotlight functions exist in the iframe), `wait_for`, `scroll_to`, the terminal card, evidence capture (`_capture_page` reads the app frame's aria/url/title), and the failure dump. `goto()` navigates the **iframe**; the wrapper document never navigates.
- **Cursor**: the dot lives in the wrapper document and rides over the iframe. No wrapper listener can hear a mousemove targeted inside the iframe, so the recorder drives the dot explicitly, one update per pointer step — the dot is where the pointer was commanded to be by construction, which closes the #186/#202 synthetic-mousemove class on this path. Coordinates need no manual offset: Playwright's `bounding_box()` answers in main-frame coordinates even for iframe elements.
- **Frame refusal**: an app answering with `X-Frame-Options` or CSP `frame-ancestors` makes Chromium cancel the iframe navigation (`ERR_BLOCKED_BY_RESPONSE`); the take refuses with a `RuntimeError` naming the header (read back via `context.request`), instead of recording a silently blank window. Both header forms exercised live (below).

## rec.page / rec.app

`rec.page` stays the wrapper `Page` — the whole browser surface, chrome included, so the escape hatch loses nothing.
`rec.app` is new: the app iframe's `Frame` on a wrapper take, and the page's main frame otherwise, so a storyboard written against `rec.app` records identically on both paths.
SKILL.md gets one table row (the file is at its line budget; the row states the flag, the split, and the refusal).

## caption_lost on this path

`domcontentloaded` fires for the main frame only, and the wrapper document is never replaced — so `_note_document_replaced` cannot fire mid-take and no `caption_lost` issue is recorded.
That is the truth, not a suppression: the caption lives in the wrapper document, a mid-take `goto()` navigates only the iframe, and the line really is still on screen — the beats that keep reporting it are right.
The #134 class ends for wrapper takes.

## Acceptance

- Wrapper take against `tests/fixture` drives `click`, `type_into`, `scroll_to`, `spotlight`, `caption` through the frame; the storyboard asserts each verb's post-condition against `rec.app` (`#status` advances on click, the table filters on type). Green in the new `tests/smoke --wrapper-only` arm (~36 s, reachable from `--web-only`, **not** in `EXPENSIVE_ARMS`, so it runs under `--cheap` on every push — `tests/unit`'s `CheapArm` sweep confirms).
- Caption band disjoint from the app rect, twice:
  - **geometry** — `WrapperChrome` in `tests/unit` asserts band∩app = ∅ (band directly below the slot, inside the window, inside the viewport) at four viewports, on `chrome_geometry` itself;
  - **pixels** — `check_wrapper_band` sweeps the band rect of the recorded `demo.mp4` (tests/_pixels primitives): the caption lights the band (contrast 0.4–3.5 empty vs ~30 lit; bars 6/12) and the app rect moves ≤ 3.0 mean luma across the caption's appearance (measured 0.65–0.79 healthy; the planted overlap defect reads 55.27).
- Frame refusal seen red, live.

## The frame-refusal demo (seen red)

`tests/fixture` served with `X-Frame-Options: DENY`:

```
RAISED: RuntimeError
MESSAGE: this app refuses to be framed: http://127.0.0.1:45447/ answered with X-Frame-Options: DENY, so Chromium blocked the wrapper take's iframe. Recording on would produce a silently blank window presented as a demo, so the take refuses instead. Record without wrapper=True, or serve the demo target without that header.
```

And with `Content-Security-Policy: frame-ancestors 'none'`:

```
MESSAGE: this app refuses to be framed: http://127.0.0.1:45377/ answered with Content-Security-Policy: frame-ancestors 'none', so Chromium blocked the wrapper take's iframe. ...
```

The refusal take records with `strict=False` on purpose: the blocked navigation also logs a console error, so under strict a swallowed refusal would still surface as `StrictTakeFailed` and the completed-without-raising assertion could never fail for the reason it names. The refusal must stand without strict's net — the first run of the injection manifest caught exactly this (the "silently blank" entry went unnoticed under `strict=True`), and the fix is in this diff.

## Review round (FIX-FIRST) — both blocking findings fixed

**1. `tests/lint` red on the branch** — chrome.py's docstring named `tests/unit` and `tests/smoke` as bare paths; the pointer sweep rightly flagged them as repo-only. Reworded to "this repository's …"; `./tests/lint` exits 0.

**2. The band silently mangled long captions** — reproduced exactly as the verifier measured it (174-char caption, bubble 130 px vs band 96, ~17 px shaved off the first and last lines, timeline clean). Fixed by recording, not by reflowing: the wrapper's `__demoCaption` now returns the pixels the band could not show (`scrollHeight − band.clientHeight`), and `caption()` records a **`caption_clipped` issue** naming the caption and the overflow, attributed to the caption beat. Not in `STRICT_KINDS` — like `caption_lost`, it is the storyboard's mistake, not the app's. The text is deliberately never capped or reflowed.

Seen green on the reviewer's reproduction (174-char caption through a real wrapper take):

```
demo-video: 1 problem(s) recorded during this take (caption_clipped x1)
  [3.00s] caption_clipped in beat 3 (caption): the caption "This caption is deliberately far too long for the reserved caption band below the app rect, so the band's edges shave its first and its last line instead of covering the app." is 34px taller than the caption band, so the band's edges shave its first and last lines and the frames do not show the sentence the beat log records — shorten the line, or split it over two captions
```

with `"clipped_px": 34` on the issue record (the verifier's 130.08 − 96). The short caption in the same take records nothing, and the strict take still completes.

Seen red, exactly-once by the harnesses: the smoke arm now captions one line the band holds and one it cannot, and `check_wrapper_clipped` requires exactly one `caption_clipped` naming the long line (plus the over-reporting control):

```
PASS  break: the band shaves a too-tall caption and no artifact says so  [29s]
      --wrapper-only, in chrome.py: if (!text) return 0; ...
      caught: wrapper: the 174-char caption left no caption_clipped issue — the band shaved its first and last lines and no artifact says so, ...
```

and two browser-free entries in `tests/unit --fault-inject`: the recorder guard neutered (`caption_clipped` never recorded → the overflow test red) and the over-fire direction (`<= 0` → `< 0`, an exactly-fitting caption recorded as clipped → the fits-test red).

## Injections — every new assertion seen red, exactly-once by the harness

`tests/smoke-inject --arm wrapper` (6 new entries — 5 from the first round plus the caption_clipped one from the review round; the harness aborts unless each search string matches exactly once, and runs the arm clean first):

```
BASE  --wrapper-only passes clean (21s)
PASS  break: the wrapper caption is dispatched and never painted in the band
      caught: wrapper: no frame of the caption band (128, 570, 1024, 96) ever reaches 12.0 contrast across 12.5s — the caption ... never painted in its band
PASS  break: the caption's appearance repaints pixels inside the app rect
      caught: wrapper: the app rect moved 55.27 mean luma between 0.40s and 1.20s, over the 3.0 bar — the caption's appearance changed pixels inside the app rect
PASS  break: evidence on a wrapper take reads the chrome, not the app
      caught: wrapper: the caption beat's evidence aria never mentions the fixture's own heading — evidence reads the wrapper chrome, not the app
PASS  break: a frame-refused app fails with an error naming no header
      caught: wrapper-refused: the refusal does not name the header that caused it — a storyboard author cannot act on Error: 'Frame.goto: net::ERR_BLOCKED_BY_RESPO...
PASS  break: a frame-refused app is recorded as a silently blank demo
      caught: wrapper-refused: recording against X-Frame-Options: DENY completed without raising — the take recorded a silently blank window as a demo
All 5 injections were caught. [3.3 min]
```

`tests/unit --fault-inject` carries 2 new entries for the browser-free half (band geometry raised into the app rect → `WrapperChrome` red; the 34px compensation inherited by the wrapper → the caption-font test red):

```
PASS  break: the caption band is raised into the app rect it must sit below
      in chrome.py: "bandy": winy + bar + pad + apph,…
      FAIL: test_the_caption_band_and_the_app_rect_share_no_pixels (WrapperChrome)
      FAIL: test_the_band_sits_directly_below_the_slot_inside_the_window (WrapperChrome)

PASS  break: a wrapper take inherits the composite's 34px caption compensation
      in web.py: if not self._wrapper:…
      FAIL: test_a_wrapper_take_keeps_the_base_caption_font (WrapperChrome)

All 347 injections were caught.
```

One pre-existing entry in each harness pinned web.py's old single-line `from .core import ...`; both are updated to the reshaped import (the unit driver's exactly-once refusal is what caught it: `ABORT ... matched 0 times`, exit 3).

## Gate (re-run after the review fixes)

- `tests/unit` — 536 tests OK (8 new in `WrapperChrome`)
- `tests/unit --fault-inject` — all 347 caught (4 new entries)
- `tests/ci-unit` — OK
- `tests/lint` — exit 0 (ruff 0.16.1; the chrome.py pointer finding fixed)
- `mypy skills/demo-video/helpers/` at 1.18.2 — no issues, 14 files
- `tests/smoke --wrapper-only` — PASSED (28–36 s measured across runs)
- `tests/smoke-inject --arm wrapper` — 6/6 caught over a clean baseline
- `tests/smoke --cheap` — PASSED (the wrapper phase runs inside it). A first run failed in the **segments** phase only, each failure naming a −914 ms host wall-clock step at 1.8 s — the #255/#263 stepping-WSL2-host class this box hits regularly (takes in this session measured −631/−840/−842/−914 ms steps); `--segments-only` and the full `--cheap` both pass clean on the identical tree.
- `tests/smoke-inject --self-test` — every guard refused what it exists to refuse (README figures updated: 70 entries, 14 arms; one pre-existing entry's search string updated for the reshaped `web.py` import)
- `tests/pixel` — re-recorded both cached takes (this diff changes `demo_recording/` bytes, so the digest invalidation is the designed behaviour) and all six checks stay green on the **legacy-path** takes: `pixel: 6 checks, 0 failing`

## Stated limits (owned by later slices, or deliberate)

- **#360 owns**: interlude/criterion cards in the wrapper (today they still raise in the wrapper *document* as full-viewport overlays — they work, but cover the chrome too and keep the #291 compensated colour), the opening hold (a wrapper take's opening gap is measured and warned about but not covered; `_opening_held` stays `None`, the terminal recorder's honest answer), and caption survival across a real `goto` graded end to end.
- **#361 owns**: the cutover, composite deletion, flag removal, geometry-consumer migration, and the encode-time saving measurement.
- **#362 owns**: mounting the terminal in `#__chrome_slot` and retiring `_TERM_HOST_JS`'s copy of the window; until then the gradient/title-bar constants exist in `chrome.py` and in the two legacy copies, byte-identical.
- The app viewport on a wrapper take is `int(W·0.8) × int(H·2/3)` (1024×480 at 1280×720) — shorter than the composite's 1024×576, because the band's 96 px comes out of the window rather than growing it. A 16:9-assuming app gets an honest 2.13:1 viewport; revisit at #361 if it matters in practice.
- The caption band holds two lines at 26px; a longer caption clips at the band's edge (`overflow: hidden`) rather than growing over the app — growing upward is the overlap this slice exists to remove. The clip is no longer silent: the take records a `caption_clipped` issue naming the caption and the overflow in px (review fix above).
- On a wrapper take, escape-hatch pointer work through `rec.page.mouse` moves the real pointer but not the cursor dot — the dot is verb-driven by design (nothing listens for events), so raw mouse moves are invisible in the frames. Wrap such work in the verbs, or accept a dot that stays parked. Stated in SKILL.md's wrapper row too.
- `_content_rect` still trims `CONTENT_CAPTION_TRIM` (20%) off the app rect's bottom on the wrapper path, although no caption can sit there — the picture check scores a slightly smaller region than it could. Conservative, not a lie; #361 can drop the trim when the composite goes.
- `shot()` stills on a wrapper take are the framed wrapper page, not full-bleed — the "stills are full-bleed, video is windowed" mismatch the design record wanted gone ends on this path, but committed guides that expect full-bleed stills will see the window.
- `terminal()`'s card raises inside the app frame (clipped to the app rect), matching where it landed on composite takes.
- Frame-refusal detection keys on Chromium's `ERR_BLOCKED_BY_RESPONSE` plus one follow-up `context.request.get` to name the header; a proxy that strips the header from the probe answer would yield the refusal with the fallback "(the follow-up request could not read which)" wording — still a refusal, never a blank recording.

Fixes #358.

### DIFF
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -237,7 +237,7 @@ three it is not the variable lowercased, so do not infer it:
 | `DEMO_VIDEO_LOCALE` | `locale` — browser locale, always applied | `en-US` |
 | `DEMO_VIDEO_SPEECH` | `speech` — force narration on/off (`1`/`0`). **`speech=False` records silently even with a key set**; with no key it is off already, and forcing it *on* without one refuses the take | auto by API key |
 | `DEMO_VIDEO_STRICT` | `strict` — fail the take on console errors / non-zero exits (`1`/`0`) | off |
-| `DEMO_VIDEO_WRAPPER` | `wrapper` — record the web app in an iframe on a recorder-owned page: chrome in the recording, caption in its own band below the app, no exit composite. `rec.page` stays the wrapper page; `rec.app` is the app's document. Refuses apps sending `X-Frame-Options`/CSP `frame-ancestors`. Transitional (#358; #361 makes it the default) | off |
+| `DEMO_VIDEO_WRAPPER` | `wrapper` — record the web app in an iframe on a recorder-owned page: chrome in the recording, caption in its own band below the app, no exit composite. `rec.page` stays the wrapper page; `rec.app` is the app's document. The cursor dot is verb-driven, so raw `rec.page.mouse` work moves it nowhere; a caption taller than its band is clipped and recorded as a `caption_clipped` issue. Refuses apps sending `X-Frame-Options`/CSP `frame-ancestors`. Transitional (#358; #361 makes it the default) | off |
 | `DEMO_VIDEO_EVIDENCE` | `evidence` — write `evidence/beat-NN.json` per beat (`1`/`0`) — see [reference/review.md](reference/review.md) | **on** |
 | `DEMO_VIDEO_VOICE_ID` | `voice_id` — ElevenLabs voice | Sarah (premade) |
 | `DEMO_VIDEO_SPEECH_MODEL` | `speech_model` — ElevenLabs model | `eleven_multilingual_v2` |
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -237,6 +237,7 @@ three it is not the variable lowercased, so do not infer it:
 | `DEMO_VIDEO_LOCALE` | `locale` — browser locale, always applied | `en-US` |
 | `DEMO_VIDEO_SPEECH` | `speech` — force narration on/off (`1`/`0`). **`speech=False` records silently even with a key set**; with no key it is off already, and forcing it *on* without one refuses the take | auto by API key |
 | `DEMO_VIDEO_STRICT` | `strict` — fail the take on console errors / non-zero exits (`1`/`0`) | off |
+| `DEMO_VIDEO_WRAPPER` | `wrapper` — record the web app in an iframe on a recorder-owned page: chrome in the recording, caption in its own band below the app, no exit composite. `rec.page` stays the wrapper page; `rec.app` is the app's document. Refuses apps sending `X-Frame-Options`/CSP `frame-ancestors`. Transitional (#358; #361 makes it the default) | off |
 | `DEMO_VIDEO_EVIDENCE` | `evidence` — write `evidence/beat-NN.json` per beat (`1`/`0`) — see [reference/review.md](reference/review.md) | **on** |
 | `DEMO_VIDEO_VOICE_ID` | `voice_id` — ElevenLabs voice | Sarah (premade) |
 | `DEMO_VIDEO_SPEECH_MODEL` | `speech_model` — ElevenLabs model | `eleven_multilingual_v2` |
--- a/skills/demo-video/helpers/demo_recording/chrome.py
+++ b/skills/demo-video/helpers/demo_recording/chrome.py
@@ -1,0 +1,244 @@
+"""The wrapper take's window chrome — one source for the frame both media share.
+
+Issue #358 (design record: #355). A wrapper take records a recorder-owned
+page carrying the window chrome: the pastel background, the dark rounded
+window with its title bar and traffic lights, a **content slot** the medium
+fills (the web recorder mounts the app iframe in it; #362 mounts xterm.js in
+the same slot), a **caption band** reserved BELOW the slot, and an empty card
+layer #360 will use. By construction the slot and the band share no pixels —
+`chrome_geometry` is the arithmetic behind that sentence, and `tests/unit`
+grades it while `tests/smoke --wrapper-only` reads the same claim out of a
+recorded take's frames.
+
+Every visual constant here is ported from `web._FRAME_HTML` and
+`terminal._TERM_HOST_JS`, which stay untouched while the legacy composite
+path exists; #361 and #362 retire those copies. The window body colour is
+**not** declared here — it stays `core.WEB_WINDOW_BODY`, because the interlude
+card's compensated colour (`core.WEB_CARD_BODY`, #291/#301) is documented
+against it and the two must be read together.
+"""
+
+from __future__ import annotations
+
+# The pastel gradient behind the window. The same string web.py and
+# terminal.py paint today (`_WEB_BG`, `_TERM_BG`); this copy is the one the
+# wrapper path reads, and the one that survives #361/#362.
+CHROME_BG = "linear-gradient(135deg, #f6d5f0 0%, #d7e3fb 52%, #cdeede 100%)"
+
+# Title bar: height, fill and text colour, and the three traffic lights —
+# ported from `_FRAME_HTML` (web) which `_TERM_HOST_JS` (terminal) matches.
+CHROME_BAR_PX = 36
+CHROME_BAR_BG = "#232334"
+CHROME_BAR_FG = "#9399b2"
+TRAFFIC_LIGHTS = ("#ff5f57", "#febc2e", "#28c840")
+
+# The pad the window keeps around the content slot, so its rounded corners
+# stay visible — `_frame_geometry`'s 14, ported.
+CHROME_PAD_PX = 14
+
+# The caption band reserved below the content slot, inside the window. Sized
+# for two lines of the wrapper caption font (2 x 26px x 1.36 line height +
+# the bubble's 24px vertical padding = 95): a caption that needs a third line
+# is clipped by the band rather than allowed to grow over the app — the whole
+# point of the band is that the app never shares a pixel with the caption.
+CAPTION_BAND_PX = 96
+
+# The wrapper caption's font size. The composite path renders captions at
+# 34px because ffmpeg scales the page by ~0.8 into the window (~27px on
+# screen); the wrapper page is recorded at true pixel size, so the caption
+# uses the base 26px directly — the same effective size the terminal
+# recorder's captions have. No compensation, because there is no scale.
+CAPTION_FONT_PX = 26
+
+# Element ids. The slot is what a medium fills; the band holds the caption
+# element; the card layer is #360's mount point and ships empty this slice.
+# The caption element deliberately keeps the id `core._CAPTION_JS` uses, so
+# "the caption element" is one selector whichever path drew it. The cursor
+# dot keeps `web._CURSOR_JS`'s id for the same reason.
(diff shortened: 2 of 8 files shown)


rogvid:
Verified in two rounds per the contract. Round 1 found two blockers: a lint-red docstring and the caption band silently shaving long captions while the timeline recorded the full text. Round 2 spot-check: both fixed — lint exits 0, and caption_clipped now writes the shave into the take's issues (seen red via the wrapper smoke-inject entry, clipped_px matches the verifier's measured 130.08−96). Non-blocking limits are stated in the body; the wrapper cursor-golden migration is pinned on #361. Merging.

Issue #163, opened 2026-07-30
The content check is anti-correlated on a full-viewport overlay: it scored the covered take 23% HIGHER than the clean one


rogvid:
Found alongside #162, and it is the more serious half. #162 is a bug an author
hits once; **this is the check that exists to catch it, scoring it backwards.**

## The measurement

Three takes of the same storyboard against the same app, from a clean
standalone install:

| take | state | `content` score |
|---|---|---|
| 1 | scrim + stale label over the app for the last ~17 s of 48 s | **32.94** |
| 2 | same, cleared with the documented call (#162) — still covered | **32.95** |
| 3 | clean, nothing covering the app | **26.74** |

**The two broken takes scored ~23% higher than the correct one**, and stderr
printed `demo.mp4 shows a picture` for all three. The scrim is a soft gradient;
gradient across the measured rect adds luma variance, and luma standard
deviation is what the content axis measures.

## Why this is the catalogue's own failure mode

`wip/verified-review/SKILL.md` lists **the anti-correlated metric** as a
measurement that grades nothing — and this repo already learned it once. From
`tests/smoke`'s own docstring, on why the *whole frame* is not scored:

> a fifth to a third of every frame is the recorder's own chrome … which on its
> own scores 60-79, so a blank recording can outscore a healthy one.

The fix then was to measure the app's rect instead of the whole frame. This is
the same defect one level in: **an occluder that sits inside the app rect and
has internal variance is scored as picture.** Restricting the rect cannot help,
because the overlay is exactly where the app is.

`SKILL.md:504` names this scenario explicitly and sends the reader to `content`:

> A card left up, a modal that never closed … produces a complete, correct,
> entirely successful timeline over a recording nobody can watch

`content` is the thing that missed it.

## What would actually see it

Not a threshold change — no floor separates 26.74 from 32.94 in the right
direction. It needs a different question. Candidates, unranked:

- **Compare against the app's own painted content**, not against absolute
  variance: an overlay that occludes makes the frame *differ from the frames
  before the overlay appeared* in a sustained way. The held-picture arm already
  reasons about sustained sameness; this is sustained *difference* followed by
  sameness.
- **Ask the page**, not the pixels. The recorder controls the overlay elements
  (`__demoInterlude` / `__demoBridge`); a checkpoint could assert at teardown
  that neither is up, which is cheap and exact for the recorder's *own* chrome.
  It would not catch an app-level modal, but it would have caught this.
[...]

PR #367, opened 2026-08-22
Wrapper chrome parity: cards over the app rect, a frame-0 opening hold, captions that survive navigation, uncompensated colour (#360)

### LINKED ISSUE #360: Wrapper chrome parity: cards and opening hold in the wrapper, captions that survive navigation, uncompensated colour

rogvid:
Slice 2 of #355's structural half. Needs #358 (merged as fc32e01). Fixes #360.

## Acceptance, per the issue's three bullets

**1. Cards render in the wrapper document, over the app rect.** chrome.py's card layer now sits exactly on the content slot and carries the interlude/criterion card and the bridge scrim, under core's own element ids — so `interlude("")`, `criterion()` and the end-of-take overlay probe (#163) work with no wrapper dispatch. The viewer keeps the window frame and the narration line during a card: the card replaces the app's content, the chrome is the recorder's. The opening hold is wrapper-native: opaque in frame 0 (the `_OPENING_CARD_JS` init-script pattern, plus a markup copy in the chrome document — measured, the init script's re-run on the `set_content` document lands only after its first paint), cleared by the storyboard's first `goto()`. The ffmpeg `enable='lt(t,held)'` overlay does not run on the wrapper path. The slot's underlay is the browser's white canvas, which is what makes "the hold was up in frame 0" falsifiable in pixels — a transparent iframe borrowing the window colour would read the same with or without the hold.

**2. The caption survives a real mid-take goto, in pixels.** The wrapper arm captions a line, sends the app frame to a new fixture page (`tests/fixture/second.html`, a deliberately mid-luma document the check locates in the app rect's own pixels), and reads the band lit before the load and at +0.2/+0.5/+0.8 s after it. Beat-log honesty is structural: the wrapper recorder does not even subscribe the `domcontentloaded` signal that `caption_lost` rides, the check asserts zero `caption_lost` issues, and the goto beat still reports the line. SKILL.md's caption row and reference/failures.md's `caption_lost` row now say so (SKILL.md edited in place — budget stays 630/630).

**3. Colour, single-encoder.** The wrapper card paints from the same `window_body` substitution as the window itself — `core.WEB_WINDOW_BODY`, one declaration, no compensation. `check_wrapper_card` reads the encoded mp4: card strip vs window pad at worst-channel |Δ| ≤ 2, measured **0.1–0.6** across runs, with a no-card control that must sit ≥ 12 off the body (reads 213). No assertion anywhere pins the declarations equal or apart (#297's recorded anti-patterns) — the check reads pixels on both sides. `WEB_CARD_BODY` and `INTERLUDE_CSS_WEB` are untouched, legacy-only, and #361 deletes them.

## Green, with numbers

```
wrapper-opening: frame 0 reads 24.0 over (230, 118, 819, 57) (hold <= 60.0), and the app shows from 0.7s (bare >= 150.0)
wrapper-card: card and window body sit 0.5 apart, worst channel over 2 samples (bar 2.0); the no-card control reads 213 off the body (premise bar 12.0)
wrapper-survives: second.html arrives at 25.9s, band lit before and at +(0.2, 0.5, 0.8)s after; no caption_lost recorded, goto beat still reports the line
wrapper: band lit 9.7s from 1.7s, app rect moved 0.08 (bar 3.0) across the caption's appearance
smoke: PASSED
```

- `./tests/unit` — 543 tests OK (8 new WrapperChrome tests)
- `./tests/unit --fault-inject` — **All 355 injections were caught** (8 new entries: card layer off the slot, card painted from a literal, hold token unsubstituted, hold script unregistered, hold never cleared / re-cleared / cleared on legacy, `caption_lost` listener re-subscribed)
- `./tests/ci-unit`, `./tests/lint`, mypy 1.18.2 per ci.yml — clean
- `./tests/pixel` — re-recorded from these source edits, **6 checks, 0 failing** (legacy goldens green; wrapper goldens are #361's, per the note pinned there)
- `./tests/smoke --wrapper-only` — three consecutive green runs; `./tests/smoke --cheap` — PASSED
- `./tests/smoke-inject --self-test` — every guard holds; README manifest figures re-synced (73 entries, 14 arms; `--wrapper-only` 9 entries at a re-measured 48 s)

## Red demos (exactly-once injections, failure output verbatim)

**Card layer transparent at frame 0** (`chrome.py`, hold `opacity: 1` → `0`):

```
BASE  --wrapper-only passes clean (50s)
PASS  break: the card layer ships transparent at frame 0  [49s]
      caught: wrapper-opening: the wrapper take's first frame reads 255.0 mean luma over (230, 118, 819, 57), over the 60.0 bar for the opening hold — the card layer was not opaque in frame 0 ...
```

**Card palette nudged 4 levels** (`chrome.py`, `background: __WINBG__` → `#1c1c29`):

```
BASE  --wrapper-only passes clean (49s)
PASS  break: the wrapper card's palette is nudged four levels off the window  [49s]
      caught: wrapper-card: 30% through the criterion beat the card reads (27, 27, 39) ... and the window body (22, 22, 34) ...
      caught: wrapper-card: 60% through the criterion beat the card reads (27, 27, 39) ... and the window body (22, 22, 34) ...
```

**Caption band killed on navigation** (`web.py`, goto clears `__demoCaption` on the wrapper path):

```
BASE  --wrapper-only passes clean (49s)
PASS  break: a mid-take goto kills the caption band  [50s]
      caught: wrapper-survives: 0.8s after the second document arrived the caption band reads 0.6 contrast, under the 12.0 lit bar — the caption band went dark across the load ...
```

All three staged copies confirmed the injection landed (exactly-once match, enforced by the harness).

## Decisions the issue left to the author

- **The decorative terminal prop stays app-side.** It is content — a prop inside the demo's story that rides over the app the way an app dialog would — and the card layer stays reserved for recorder furniture. #362 mounts a *real* terminal in the slot, and a prop on the chrome layer would blur exactly that line. (chrome.py module docstring states this.)
- **The hold clears at the end of the first `goto()`**, inside that beat — the storyboard's first content beat, the moment there is an app to reveal.

## Stated limits (#361/#362 boundaries and other measured edges)

- `content.opening.gap` on a wrapper take describes the hold's own featureless stretch (~0.7 s here) with `held: null` — a description, warning-free, since only a recorder that claims to cover can warn. Whether wrapper takes should self-report the hold the way terminal takes report their card (#235's `content.opening.card`, still `null` on web) is #361's reporting sweep.
- A wrapper storyboard that never calls `goto()` keeps the hold up for the whole take; the picture check reads a blank app rect and `check_wrapper_opening`'s bare-frame control goes red in the arm, but no recorder-side warning names the hold specifically.
- The bridge scrim renders in the card layer and is graded structurally (ids, unit tests); no pixel check reads the scrim itself on the wrapper path — the legacy overlay arm still grades scrim behaviour.
- Wrapper card text sizes match the terminal recorder's declared 30 px/34 px (the shared chrome #362 unifies on), not the composite's ~0.8-scaled effective sizes.
- Legacy path byte-untouched in behaviour: web.py's composite, `_opening_hold`, `WEB_CARD_BODY`, `INTERLUDE_CSS_WEB` all remain; #361 owns the cutover and deletions, #362 the terminal mount.

### DIFF
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -264,7 +264,7 @@ exception, or a non-zero exit. Both are
 |---|---|
 | `goto(path)` | Navigate (relative to base_url); waits for networkidle, but gives up after 10 s for apps that poll |
 | `pause(s)` / `shot(name)` | Hold the frame / capture `images/<name>.png` |
-| `caption(text)` | Narrator line at the bottom; `""` clears; dies on full page loads — the beat log clears with it and records a `caption_lost` issue naming the line — survives SPA routing; re-caption after a load |
+| `caption(text)` | Narrator line at the bottom; `""` clears. On the composite path it dies on full page loads — the beat log clears with it, records a `caption_lost` issue naming the line, survives SPA routing; re-caption after a load. A wrapper take's caption lives in the recorder's own band, survives navigation, and `caption_lost` can never fire there |
 | `caption(text, ac="AC-3")` / `shot(name, ac="AC-3")` | Tag this beat with the acceptance criterion it is there to demonstrate. Needs `Recorder(criteria={...})`; a tag naming an undeclared criterion is refused. See [reference/review.md](reference/review.md). |
 | `criterion("AC-3")` | Raise a card carrying **AC-3's own declared sentence**, read out of `criteria={...}` rather than retyped — so the viewer meets the clause and then watches it happen. The beat claims AC-3 and nothing else; the beats after it are untagged. Held to reading speed, and cleared by `interlude("")` like any card. |
 | `hold(min_s=1.5)` | Keep the current frame up until the current caption's narration finishes (min `min_s`). Use after a spotlight/action so the emphasis rides the whole spoken line instead of flashing. See **Pacing and perception** below. |
--- a/skills/demo-video/helpers/demo_recording/chrome.py
+++ b/skills/demo-video/helpers/demo_recording/chrome.py
@@ -4,19 +4,38 @@ Issue #358 (design record: #355). A wrapper take records a recorder-owned
 page carrying the window chrome: the pastel background, the dark rounded
 window with its title bar and traffic lights, a **content slot** the medium
 fills (the web recorder mounts the app iframe in it; #362 mounts xterm.js in
-the same slot), a **caption band** reserved BELOW the slot, and an empty card
-layer #360 will use. By construction the slot and the band share no pixels —
-`chrome_geometry` is the arithmetic behind that sentence. This repository's
-`tests/unit` (not shipped with the installed skill) grades it, and this
-repository's `tests/smoke --wrapper-only` reads the same claim out of a
-recorded take's frames.
+the same slot), a **caption band** reserved BELOW the slot, and a **card
+layer** over the slot (#360). By construction the slot and the band share no
+pixels — `chrome_geometry` is the arithmetic behind that sentence. This
+repository's `tests/unit` (not shipped with the installed skill) grades it,
+and this repository's `tests/smoke --wrapper-only` reads the same claim out
+of a recorded take's frames.
 
-Every visual constant here is ported from `web._FRAME_HTML` and
+**The card layer covers the app rect — not the chrome, not the caption
+band.** `interlude()`, `criterion()` and the `light` bridge scrim render in
+it, so a card replaces the *app's* content and nothing else: the window
+frame and the narration line are the recorder's own furniture, and taking
+them down to show a recorder-authored card would be the chrome hiding from
+itself. The legacy composite path cannot make that distinction — its card is
+an element inside the app page, so full-bleed there means the whole page —
+which is why its card sits above the caption bar and this one sits beside
+it. The decorative terminal prop (`Recorder.terminal()`) is deliberately
+**not** here: it is content, a styled element inside the demo's story that
+rides over the app the way an app dialog would, and the card layer stays
+reserved for the recorder's own furniture — #362 mounts a *real* terminal in
+the slot, and a prop living on the chrome layer would blur exactly that line.
+
+**The wrapper card declares the window's own colour — no compensation.** The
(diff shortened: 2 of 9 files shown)


rogvid:
Verified against the two-question contract: all three acceptance bullets met in pixels (frame 0 reads 24.0 under the 60 bar; card vs window 0.3 worst-channel uncompensated with a 213-level control; caption survives the goto to second.html), legacy path untouched (core.py not in the diff, every change behind the wrapper flag). The verifier re-ran the transparent-hold injection itself: frame 0 read 255.0 — the white slot underlay is what makes the claim falsifiable. Non-blocking notes stay in the PR body; the first-goto beat's card-showing review frame is inherited from legacy behaviour and lands in #361's reporting sweep. Merging.

Issue #291, opened 2026-08-12
On a web take, the interlude card is painted the same colour as the skill's terminal widgets and reads as a terminal


rogvid:
**Found by a human watching the `criterion()` demo (#285)** — not by any assertion, and no assertion here could have found it. The aria snapshot carries the right text, the frames are correct, `timeline.json` is correct, and the take is a `WebRecorder` take from first beat to last. Same class as #110/#111, which the commit that fixed them called "the first defects in this project caught by eye rather than by a test or a code reviewer".

### What a viewer sees

`examples/ticket-queue/demos/2026-08-11-criterion-card/frames/beat-06.png`: the recorder's own window frame (`web.py:39-53` — dark rounded window, title bar, traffic-light dots, `127.0.0.1:8901`) with the interlude card filling the entire viewport inside it. The app is gone, the frame is a flat dark field with centred text, and the result reads as **a terminal window**, not as a card over a web app.

### Why

`INTERLUDE_CSS` (`core.py:118-124`) paints `background: #1c1a17` across `inset: 0`. That is the same value as `__demoTerminal`'s card background at `web.py:213` — the skill's own terminal-styled widget. So the card does not read as an interlude over the app; it reads as the app having been replaced by a terminal.

The palette match is not an accident. `eca42c5` ("Ease the spotlight out, and **open a terminal segment on its card**", #110/#111) made the interlude card the thing a *terminal* segment opens on, to kill a ~290 ms flash of bare terminal. Tuned to blend with a terminal, it blends with a terminal — and on a **web** take that is the wrong thing to blend with.

### Why it surfaced now

Nothing in #285 touches `INTERLUDE_CSS` (`git log -S` shows `eca42c5` as its last change) and `criterion()` reuses `interlude()`'s element verbatim. What changed is dwell: `criterion()` holds at reading speed, so the demo's two cards are up for 6.73 s and 7.75 s — 15 s of a 58 s recording. A pre-existing ambiguity that used to flash past in `interlude()`'s flat 2.8 s is now on screen long enough to be misread.

### Acceptance criterion

A viewer of a web take can tell a card from a terminal. **Verified by watching** — this is demoable by the repo's own test, so the fix gets a demo and a human reviews it.

[...]

Issue #263, opened 2026-08-10
_merge_timeline can overlap two beats: part 2 is offset by ffprobe duration while part 1's log is monotonic


rogvid:
Found while characterising #259; **reproduced in a constructed stitched take**, and it is a third mechanism, distinct from #255's hole and not in `capture_clock` at all.

`stitching.py`'s `_merge_timeline` offsets part 2 by part 1's **ffprobe duration**, while part 1's beats keep their own monotonic timestamps. When part 1's clock stepped, those two disagree, and the merged log goes non-monotonic:

```
beat 21  t_end   37.451
beat 22  t_start 37.441      <- starts before the previous beat ended
```

from a −1.4684 s step. The overlap scales with Δ minus startup and tail, so on this host's ~1.5 s steps it can reach most of a second.

This is what PR #261 saw in its discarded take and could not explain.

## Acceptance criterion

A stitched timeline is monotonic, or says why it is not.

- [ ] The merge either clamps the overlap or reports it — silently emitting a non-monotonic log is the one option that is not available.
- [ ] Whichever is chosen is visible to a reader of `timeline.json` / `timeline.md`, not only on stderr.
- [ ] Graded by a constructed case: the host cannot be relied on to step to order.

## Deliberately out of scope

#255's hole (fixed at the source in #256) and #259's verification pass. This is a merge-arithmetic defect and should be fixed as one.

PR #368, opened 2026-08-23
The wrapper is the only web path: composite, compensation and the event-following cursor deleted

### LINKED ISSUE #361: Wrapper cutover: one web path, composite deleted, consumers and docs moved, encode saving measured

rogvid:
Slice 3 of #355's structural half (design record there). Built on #358 (fc32e01) and #360 (5a40204). Fixes #361.

## Acceptance, per the issue

- **The wrapper is the only web path.** The transitional `wrapper=` flag and `DEMO_VIDEO_WRAPPER` are gone; `Recorder` records wrapper takes unconditionally. `rec.app` is always the app iframe's frame, `rec.page` the framed page.
- **The exit-time composite is deleted** — the full list below.
- **Grep acceptance**: `_postprocess` survives only as the no-op base hook (the medium seam a future transforming medium would use; both shipped recorders leave it alone); `WEB_CARD_BODY` appears only inside history notes (core.py's `WEB_WINDOW_BODY` note, web.py's module docstring, tests/README's history paragraphs). No compensation constant exists.
- **Full `tests/smoke` green on the wrapper-only path**, `--web-only` and `--cheap` included — gate below.
- **The retake measurement shows one encode saved** — table below.

## The deletion list

| deleted | where | size |
|---|---|---|
| `_FRAME_HTML` + `_WEB_BG` + the screenshot render half of `_start` | web.py | ~55 lines |
| `_postprocess` composite + `_opening_hold` + the `enable='lt(t,held)'` overlay | web.py | ~91 lines |
| `_CURSOR_JS` + the #186/#202/#203 rationale block (the event-following dot) | web.py | ~133 lines |
| `_note_document_replaced` + the web `domcontentloaded` subscription (#134/#180 machinery) | web.py | ~75 lines |
| `wrapper=` flag, `_wrapper` dispatches, the 34px caption compensation | web.py | ~40 lines |
| `WEB_CARD_BODY`, `INTERLUDE_CSS_WEB`, and the compensated-palette essays, condensed into history notes pointing at #291/#301/#355 | core.py | 113− / 57+ |
| `check_opening` (composite-hold check) + its section; `check_criterion_card` + `WEB_CARD_RGB`/`CARD_RGB_TOLERANCE`/`CARD_WINDOW_TOLERANCE`; `check_parked_pointer`; legacy `caption_band` probes | tests/smoke | 627− / 321+ |
| `CursorMotion` + the JS-extraction harness; the caption-clearing halves of `CaptionTruth`/`MeasuredNavigations`; `CaptionLost`; `InterludePalette`'s web-palette tests; 15 fault-inject entries for the deleted machinery | tests/unit | 1450− / 696+ |

Branch total: **+1,900 / −3,365** across 19 files. web.py alone: 182+/525−.

## caption_lost semantics

Web takes no longer subscribe the document-load signal at all — unconditional since this PR (`_watch_extra` keeps only #142's `framenavigated` carve-out). `caption_lost` stays a published issue kind: committed timelines carry it and `stitch()` merges old segments' issues as-is (`CaptionLostRetired` in tests/unit pins both, with two injections). SKILL.md's caption row and reference/failures.md now say caption death on navigation is a terminal-take concern only (its caption is in-page, until #362) — and a terminal take's page never navigates.

The measured `NAVIGATIONS` table (Chromium event streams, #179) stays as data; its sweep inverted: every measured way of leaving a page must now leave the caption on the beat log and file no issue, non-vacuously (every stream lands on the kept `framenavigated` listener). `tests/smoke --wrapper-only`'s check_wrapper_caption_survives keeps the pixel half.

## Geometry consumers moved

- `_content_rect` reads `chrome.chrome_geometry`'s dict as the single source.
- `to_video_rect` is now the app-rect **offset**, no scale — the ~0.8 factor died with the composite. Most call sites dropped mapping outright: `bounding_box()` answers in wrapper-page = video coordinates (spotlight rect, entropy panel/spinner rects).
- Stills and video share one geometry: `check_take` gets one rect for both; the "stills are full-bleed, video is windowed" mismatch is gone, and `shot()` stills match the video frame for frame.
- Caption probes read the chrome's band rect (`caption_probe_band`, chrome_geometry's arithmetic duplicated per the loop-vs-suite rule), not a full-width bottom strip. Floors re-measured on the band: probe caption 4.99 healthy vs 0.00 undrawn → `MIN_CAPTION_BAND_DIFF`/`MIN_ALIGN_BAND_DELTA` web/segments 8.0/6.0 → 2.0 (2.5x margin).
- Determinism/ticker probes target `rec.app` — the claims are about what the recorder does to the *app*, and the wrapper document's init-script re-run is racy on `set_content` (measured: the motion-CSS control read 0.18s in the wrapper doc on one run).

### Legacy-only checks: moved or died, one reason each

| check | fate |
|---|---|
| `check_opening` (held>0 + frame-0-vs-median) | **died** — it graded the deleted ffmpeg hold. The same take now runs `check_wrapper_opening` (frame 0 = hold ≤ 60 luma, bare app ≥ 150 later), so the wrapper opening is graded on the 28 s web take and the wrapper arm alike |
| `check_opening_gap` | **stays** — it grades `opening_gap`, the measurement `content.opening` reports on every take, on synthesised videos |
| `check_criterion_card` (compensated pair) | **died** — no pair left. The coverage take's card is graded by `check_wrapper_card`: one declaration, one encoder, pixels on both sides |
| `check_parked_pointer` (DOM == parked coordinate) | **moved** → `check_undrawn_pointer`: the determinism storyboard parks nothing (nothing but a verb can draw the dot), so it asserts the dot is off-viewport with no verb run — failing on a missing element, so it cannot go silent |
| `CaptionLost` + the caption-clearing sweeps + the #214 late-half test | **died/inverted** — the machinery is gone; see caption_lost above |
| `CursorMotion` + JS extraction | **died** — no event rule left to extract. Replaced by `CursorDrive` (one dot command per pointer step, seeded from `_cursor_at`, squeeze on press; 4 injections) and the two pixel goldens |
| `InterludePalette` web tests | **died** — no per-medium CSS dispatch. Kept: the terminal palette, the #110 opening-card agreement, and the create-if-missing guard that keeps the base stylesheet inert where the chrome ships the card element (1 new injection) |

## Pixel goldens migrated (the pinned requirement)

The web reference storyboard records a wrapper take; the marker's geometry is `chrome_geometry`'s dict, and the park is the app-document rect mapped through the app-rect offset (`to_video_rect`) — so the mapping the verbs rely on is graded rather than assumed. `card-vs-window` reads the uncompensated declaration (`WEB_WINDOW_RGB` = `core.WEB_WINDOW_BODY`), locating the card by the dark run that **overlaps the logged beat** (`card_run`) because the opening hold is a second, deliberate dark stretch in the same strip. `opening-card` (terminal) is untouched.

Seen red by the issue's two planted defects, the digest re-recording automatically both times:

**cursor offset edit** (web.py, `__demoChromeCursor(x, y)", [xi, yi]` → `[xi + 30, yi]`):

```
web/cursor-parked: FAIL - disc centre (620.5, 401.3) vs park (590.5, 401.0), off (+30.0, +0.3)px (bar 3.0), 235 accent px
```

**dot-at-load edit** (chrome.py, `#__demo_cursor { ... top: -40px; left: -40px` → `top: 8px; left: 8px`):

```
web/cursor-absence: FAIL - 9 frames before beat 11 probed, worst frame 144 accent px in the 16x16px origin crop (bar 0)
```

Both reverted; re-recorded; `pixel: 6 checks, 0 failing`. Healthy: disc centre off (+0.6, +0.1)px, card 3 off the window body (bar 5, shared with smoke — see the stated limit on what this reading resolves).

## The reporting sweep (#367's stated limit, inherited)

The first goto's review frame shows the opening hold — a flat dark window under a heading that says `goto`. The goto that clears the hold now stamps the instant on its own beat (`opening_hold_until`; `stitch()` shifts it with the beat), and `frames.md` names the frames cut inside it, in `scene_search_skipped`'s named-not-counted style. Live, from the web take:

> `beat-00.png` was cut inside the take's opening hold: until 0.63s the app rect is an opaque field in the window's own colour, up from frame 0 and cleared when the first `goto` had an app to reveal. A flat dark window under this heading is the hold, not an app that failed to load.

Seen red, exactly-once, in `tests/unit --fault-inject` (4 entries): the note dropped whole, the note sweeping in other segments' frames, the goto not stamping the instant, and stitch keeping the instant on the segment's own clock:

```
PASS  break: the sheet stops naming the frames cut inside the opening hold
      in frames.py: manifest["opening_hold"] = holds…
      FAIL: test_the_frames_cut_inside_the_hold_are_named_on_the_sheet (OpeningHoldOnTheSheet)
      FAIL: test_a_merged_sheet_scopes_each_hold_to_its_own_segment (OpeningHoldOnTheSheet)
```

## Viewer-visible changes, and one for evidence readers

- **Captions render at the base 26px.** The 34px compensation existed for the composite's ~0.8 downscale; the page records at true pixel size now. (Card text stays the declared 30/34px from #360.)
- **`goto()` takes the recorder's cards down.** A full page load used to destroy the card with the app's document, and SKILL.md's segment pattern (interlude, then goto) leans on it; the cards live in the wrapper document now, so goto clears them deliberately — every goto, as any full load did. Seen red in pixels first: `--segments-only` went red with the recorder's own overlay-left-up warning on a take the suite grades as healthy, which is exactly the take a user's segmented demo would have recorded. The caption is *not* cleared — its survival is the point.
- **Evidence gains a `chrome` field** — the visible caption line and any card, read out of the wrapper document's DOM at capture time, never copied from the beat log. The caption left the app document with the band, so without it no evidence file could say which narration line the frame showed (#9): `check_evidence`'s caption facts and the coverage arm's clause-on-screen check both read it now. Documented in reference/review.md; `EVIDENCE_LIMITS` gains `chrome: 8000`; schema stays 1 (additive key).

## Measurements

Method: the pixel loop's web reference storyboard, recorded 3x per tree against `tests/fixture` on an ephemeral port, ffmpeg shimmed to log argv (`libx264` occurrences = video encodes). Before = main (5a40204) via `git archive`; after = this branch. The after-storyboard holds one extra `pause(1.0)` (the card needs a settled app stretch), so the saving is understated by ~1 s.

| | before (composite) | after (wrapper) |
|---|---|---|
| x264 video encodes per retake | **2** (`_convert` + `_postprocess` composite; the take's 0.45 s opening hold rode the composite pass) | **1** (`_convert`) |
| wall-clock per retake | 24.4 / 23.3 / 23.5 s | 22.6 / 22.2 / 22.2 s |

## The gate

- `./tests/unit` — 534 tests OK
- `./tests/unit --fault-inject` — **All 341 injections were caught** (was 355: −21 with the deleted machinery, +7 new — the four hold-note/stamp entries, the goto card clear, the clearing-subscription-grows-back, the interlude create-guard; the four `CursorDrive` entries replace six `CursorMotion` ones)
- `./tests/ci-unit` (147 OK), `./tests/lint`, mypy 1.18.2 per ci.yml — clean
- `./tests/unit --budget` — SKILL.md 630/630 (in-place edits only)
- `./tests/pixel` — re-recorded from these source edits, 6 checks, 0 failing
- `./tests/smoke --cheap` — PASSED
- `./tests/smoke --web-only` — PASSED (the expensive arm, on the unified checks)
- `./tests/smoke --segments-only`, `--polish-only`, `--determinism-only`, `--overlay-only`, `--coverage-only`, `--evidence-only`, `--issues-only`, `--strict-only`, `--wrapper-only` — PASSED individually while unifying
- `./tests/smoke-inject --self-test` — every guard holds; README figures re-synced (71 entries, 14 arms, 16 of 49 check functions ungraded)

Four harness fixes rode along, all pre-existing stepping-host defects this box's ~1 s/32 s clock cadence surfaced while gating the branch, each with the failing run's numbers in its comment:

- `check_spotlight_transitions` now applies the suite's own wall-clock correction (#215) to its two windows — a −1.07 s step at 0.789 s of a 6 s take slid the exit window off the fade entirely; the arm was the only timing check left uncorrected.
- `_check_frame_captions`' guard compared clocks across domains: it fed the frame's *video*-clock instant into `HostClock.before()`, which takes log-clock seconds. On a take whose step landed inside a caption stretch, the corrected interval missed the change by ~1 s and a frame cut 60 ms from a caption clear was graded (beat 4 at video 3.974 s = log 5.09 s, change at log 5.151 s, step −1.119 s at 3.768 s). Each account now measures against the change on its own clock, with a straddled change reading zero.
- `check_timeline`'s monotonic sweep now honours the seam overlaps a stitched timeline *publishes* (#263 — "reported, not clamped" is the recorder's documented design): a backward step deletes wall time from a part's video without touching its log, so the next part's offset lands before the previous log's end. A seam the document reports — same beat, `seam: true`, published size covering the observed one — is the healthy shape on a stepping host; an unreported overlap, or one bigger than published, still fails, anywhere.
- The wrapper arm's survives-storyboard holds 2.5 s (was 1.0 s) between the mid-take goto and the caption clear: a −1.05 s step in that gap deleted it from the video and pulled the clear into the check's +0.5/+0.8 s samples on a healthy take.
- `check_wrapper_card` adopted the pixel golden's locate-then-measure (`card_run` moved to tests/_pixels.py, shared): it sampled at fixed fractions of the *logged* criterion span, and on a stepping host the video slides under the log, so 60% of the span landed on the card's fade — measured on a kept healthy take: gap 0.0 from 10–50%, 5.6 at 60% (mid-fade, still under the ≤ 60 locate bar), 205 at 70% (card gone). Those fade readings are what the bar-raise to 5 had explained as encoder noise. The check now finds the card's own dark stretch overlapping the logged beat, trims two sweep frames off each end, and samples inside it: the same take reads 0.1, `--web-only` went from 2 failures in 4 runs to 3/3 green, and the opacity-0 planted card is caught one gate earlier (the only overlapping dark stretch is the 0.6 s opening hold — "flashed rather than held").

## Stated limits

- **#362 owns the terminal mount and the shared caption path**: `core._CAPTION_JS`, `_caption_bottom_px` and `INTERLUDE_CSS_TERMINAL` survive for the terminal recorder only.
- **The #214 caption re-read is unexercised in effect**: a beat still pumps before stamping its inherited caption, but no page event can change the caption any more; the pump stays for event-attribution freshness. Its two fault-inject entries died with the tests that watched them (tests/README's Known gaps says so).
- **Raw `rec.page.mouse` moves the pointer and not the dot** — the escape-hatch limit inherited from #366, now the stated limit of the only path (SKILL.md, reference/determinism.md).
- **`content.opening.held` is null on every take**; the field, `limit` and `opening_warning` stay for committed timelines and any future encode-time-cover medium (content.py's history note).
- **Nothing reads frame zero of the segments or entropy takes** — they open on the same hold; only `web/` and `wrapper/` are graded (tests/README states it).
- **The card-vs-window reading resolves repaints, not nudges.** With one declaration on one encoder, the healthy card/pad pair encodes 0.5–3.8 apart (x264 settles the two rects' chroma against different block histories). Both suites' bars sit at 5, and the planted defect is an 8-level repaint (reads ~9; the terminal palette reads ≥ 12) — a nudge under ~5 declared levels is inside the encoder's own noise and is not a claim either check makes. Stated at both constants. The 5.6–5.9 readings once reported as healthy noise were not noise — see harness fix 5.
- **`check_wrapper_card` prints `wrapper-card:`-prefixed failures on the coverage arm** — the check names itself, not the arm.
- **The wrapper document's init-script re-run on `set_content` is racy** (measured: motion CSS absent from the wrapper doc in one run of three): the chrome carries its own copies of everything load-bearing (#360's pattern), and the harness probes the app document, but chrome fades can run unflattened on a deterministic take. Cosmetic; no check reads a chrome fade's duration.
- **Evidence schema stays 1** with the additive `chrome` field; pre-#361 readers see one extra key.

### DIFF
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -64,12 +64,13 @@ the web recorder; the **Terminal demos** section below covers what differs,
 and is **Unix-only** because it uses a PTY.
 
 Both record into a **framed window on a soft pastel background** (rounded
-window, title bar, traffic-light buttons) so the eye has an obvious focus —
-the caption sits as a lower-third. The terminal frames itself in-page; the
-web recording is composited into the window by ffmpeg on exit, which scales
-it down (hence the larger web caption). One consequence: web `shot()` stills
-are captured full-bleed (no window frame) — good for embedding in a guide,
-but they will not match the windowed video exactly.
+window, title bar, traffic lights), framed in-page: the web app records in
+an iframe at true pixel size, caption in a reserved band **below** the app
+rect, so the page is the finished picture — no exit composite, one encode
+per take, and `shot()` stills match the video exactly, chrome included.
+`rec.page` is the framed page; `rec.app` is the app's document. Two edges:
+an app answering `X-Frame-Options`/CSP `frame-ancestors` refuses the take
+by name, and the dot is verb-driven — raw `rec.page.mouse` never moves it.
 
 Each demo gets one folder (suggested: `docs/guides/<YYYY-MM-DD>-<slug>/`):
 
@@ -237,7 +238,6 @@ three it is not the variable lowercased, so do not infer it:
 | `DEMO_VIDEO_LOCALE` | `locale` — browser locale, always applied | `en-US` |
 | `DEMO_VIDEO_SPEECH` | `speech` — force narration on/off (`1`/`0`). **`speech=False` records silently even with a key set**; with no key it is off already, and forcing it *on* without one refuses the take | auto by API key |
 | `DEMO_VIDEO_STRICT` | `strict` — fail the take on console errors / non-zero exits (`1`/`0`) | off |
-| `DEMO_VIDEO_WRAPPER` | `wrapper` — record the web app in an iframe on a recorder-owned page: chrome in the recording, caption in its own band below the app, no exit composite. `rec.page` stays the wrapper page; `rec.app` is the app's document. The cursor dot is verb-driven, so raw `rec.page.mouse` work moves it nowhere; a caption taller than its band is clipped and recorded as a `caption_clipped` issue. Refuses apps sending `X-Frame-Options`/CSP `frame-ancestors`. Transitional (#358; #361 makes it the default) | off |
 | `DEMO_VIDEO_EVIDENCE` | `evidence` — write `evidence/beat-NN.json` per beat (`1`/`0`) — see [reference/review.md](reference/review.md) | **on** |
 | `DEMO_VIDEO_VOICE_ID` | `voice_id` — ElevenLabs voice | Sarah (premade) |
 | `DEMO_VIDEO_SPEECH_MODEL` | `speech_model` — ElevenLabs model | `eleven_multilingual_v2` |
@@ -264,7 +264,7 @@ exception, or a non-zero exit. Both are
 |---|---|
 | `goto(path)` | Navigate (relative to base_url); waits for networkidle, but gives up after 10 s for apps that poll |
 | `pause(s)` / `shot(name)` | Hold the frame / capture `images/<name>.png` |
-| `caption(text)` | Narrator line at the bottom; `""` clears. On the composite path it dies on full page loads — the beat log clears with it, records a `caption_lost` issue naming the line, survives SPA routing; re-caption after a load. A wrapper take's caption lives in the recorder's own band, survives navigation, and `caption_lost` can never fire there |
+| `caption(text)` | Narrator line in the caption band below the app; `""` clears. The band is the recorder's own document, so the line survives full page loads and SPA routing alike and `caption_lost` cannot fire on a web take — caption death on navigation is a terminal-take concern only (its caption is in-page, until #362). A line taller than the two-line band is shaved at the band's edges and recorded as a `caption_clipped` issue: shorten it, or split it over two captions |
 | `caption(text, ac="AC-3")` / `shot(name, ac="AC-3")` | Tag this beat with the acceptance criterion it is there to demonstrate. Needs `Recorder(criteria={...})`; a tag naming an undeclared criterion is refused. See [reference/review.md](reference/review.md). |
 | `criterion("AC-3")` | Raise a card carrying **AC-3's own declared sentence**, read out of `criteria={...}` rather than retyped — so the viewer meets the clause and then watches it happen. The beat claims AC-3 and nothing else; the beats after it are untagged. Held to reading speed, and cleared by `interlude("")` like any card. |
 | `hold(min_s=1.5)` | Keep the current frame up until the current caption's narration finishes (min `min_s`). Use after a spotlight/action so the emphasis rides the whole spoken line instead of flashing. See **Pacing and perception** below. |
@@ -397,11 +397,11 @@ terminal demo takes, and the gotchas — pagers, echo, prompts that never return
      `rec.spotlight(selector)` on the element it talks about, so viewers
      never see a highlight belonging to the previous line.
 3. **Caption every beat.** A fresh caption before each thing the viewer
-   should understand, a new one after any navigation, `caption("")` to end
-   clean. On a full page load the caption dies with the DOM; in an SPA
-   (hash routing) it *survives* into the next view and reads as a mistimed
-   line — so `caption("")` before the click that navigates, fresh caption
-   after, in both cases. Rules (each earned in a fresh-eyes review round):
+   should understand, `caption("")` to end clean. The caption lives in the
+   recorder's own band and survives every navigation — full loads and SPA
+   routing alike — so a line left up across one reads as narrating the next
+   view too: `caption("")` before the click that navigates, fresh caption
+   after. Rules (each earned in a fresh-eyes review round):
    - A caption may only claim what is visible in the frame. Don't promise
      an action the demo never performs — scope the words to what's shown,
      or show it. The prohibition alone does not hold — captions get written
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -581,7 +581,8 @@ terminal demo takes, and the gotchas — pagers, echo, prompts that never return
   waste the take — script 1–2 steps per message.
 - **Layout shifts strand the cursor.** Elements the app inserts mid-recording
   move, the cursor does not — re-`move_to` after any wait that can reflow.
-  The dot stays undrawn until the pointer first moves, and after each `goto()`.
+  The dot stays undrawn until the first pointer verb — then it keeps its last
+  parked spot, `goto()` included: it lives in the recorder's page, not the app.
 - **A blind click to dismiss an overlay.** A `mouse.click(400, 400)` landed
   on a file row, opened a dialog, and its backdrop swallowed every later
   click — dead take. Dismiss via a named neutral element, never coordinates.
--- a/skills/demo-video/SKILL.md
+++ b/skills/demo-video/SKILL.md
@@ -580,9 +580,8 @@ terminal demo takes, and the gotchas — pagers, echo, prompts that never return
   agent turns with a budget, one big scripted message can exhaust it and
   waste the take — script 1–2 steps per message.
 - **Layout shifts strand the cursor.** Elements the app inserts mid-recording
-  move, the cursor does not — re-`move_to` after any wait that can reflow.
-  The dot stays undrawn until the first pointer verb — then it keeps its last
-  parked spot, `goto()` included: it lives in the recorder's page, not the app.
+  move, the cursor does not — re-`move_to` after any reflowing wait. The dot
+  draws at the first pointer verb and keeps its spot across `goto()`.
 - **A blind click to dismiss an overlay.** A `mouse.click(400, 400)` landed
   on a file row, opened a dialog, and its backdrop swallowed every later
   click — dead take. Dismiss via a named neutral element, never coordinates.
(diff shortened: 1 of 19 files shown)


rogvid:
Round-1 blockers fixed in the follow-up commit: the ruff reformat, and SKILL.md's cursor sentence now states the wrapper reality (the dot keeps its last parked spot across goto() — it lives in the recorder's page). Also added to the viewer-visible list: the dot no longer resets on navigation. The verifier's non-blocking items stay in the body as stated limits.

PR #401, opened 2026-08-26
refactor(smoke): split monolithic smoke test into modular package

rogvid:
The 13,725-line `tests/smoke` monolith is now a thin executable entry over the `tests/smokekit/` package:

- `constants.py` — every threshold and take definition
- `support.py` — servers, clocks, frame/media helpers
- `checks.py` — the suite's assertion functions
- `records.py` — the take recorders
- `runs.py` — the arms `run_phases` selects between
- `cli.py` — argument handling and phase orchestration

Every line is verbatim from the pre-split file, redistributed by an AST-driven splitter that computes cross-module imports from a free-variable analysis (verified acyclic). All 420 top-level names are accounted for. `tests/smoke` stays the executable so every existing consumer — ci.yml, mise tasks, fault-inject staging, `tests/unit`'s SourceFileLoader — keeps working unchanged in shape.

**Verification** (local): `tests/lint`, `tests/typecheck`, `tests/unit` (659 OK), `tests/smoke --lock-only`, and the full `--cheap` gate all pass.


rogvid:
The first push gutted the suite (82 of 162 definitions missing, broken imports) — lint, unit and the cheap-arms job all caught it. Pushed bba982e rebuilds the split from the original source with all logic restored verbatim behind `tests/smokekit/`; details in the updated PR description. Local runs of lint / typecheck / unit / --lock-only / --cheap pass; CI should confirm.

